Issue #37, opened 2015-02-17
Correctly supervise function calls.


cmeiklejohn:
Correctly supervise operations such as `map`, `fold`, etc., to ensure they are restarted correctly.  This is inter-replica and intra-replica.

PR #44, opened 2015-05-03
Lasp "process" abstraction.

cmeiklejohn:
Properly supervise all Lasp processes through a new abstraction called
the Lasp "process."  A Lasp "process" is a process that never
terminates; it reads input values from one or more CRDTs, transforms
them, and creates output values, both streams, of which are fully
monotonic.  This is the core abstraction in Lasp, and represents a
single node in the dataflow execution graph.


cmeiklejohn:
Addresses #37.

PR #70, opened 2016-02-24
Delta mutation

junghunyoo:


apps/lasp/src/lasp_orset.erl
```diff
@@ -100,24 +100,55 @@ value(_, ORSet) ->
     value(ORSet).
 
 -spec update(orset_op(), actor(), orset()) -> {ok, orset()} |
+                                              {delta, orset()} |
                                               {error, {precondition, {not_present, member()}}}.
 update({add_by_token, Token, Elem}, _Actor, ORDict) ->
-    add_elem(Elem, Token, ORDict);
+    case application:get_env(lasp, delta_mode, true) of
```
cmeiklejohn: I would move this check to `lasp_type.erl`.  Everywhere an update is performed in Lasp, goes through `lasp_type:update`, so you can do this check once there, and have it call the delta one if the `delta_mode` is enabled.  

Make sense?

cmeiklejohn:
I merged in the changes from the `unstable` branch, so remember to pull before you begin work tomorrow!

apps/lasp/src/lasp_core.erl
```diff
@@ -337,6 +337,26 @@ bind(Id, Value, Store) ->
 
 %% @doc Define a dataflow variable to be bound a value.
 -spec bind(id(), value(), function(), store()) -> {ok, var()}.
+bind(Id, {delta, Value}, MetadataFun, Store) ->
```
cmeiklejohn: This is still going to result in the entire state being transmitted by the broadcast backend, so we need to figure out a way in the broadcast backend to only send the delta.  We probably need to change the return value from this function to return only the delta, so the broadcast backend uses that for transmission.

Make sense?

cmeiklejohn:
Additionally, there appears to be problems with the `remove_elem` function you added not handling the case of delta mutations:

```
      lasp_lattice: lasp_orset_inflation_test...*failed*
in function orddict:find/2 (orddict.erl, line 86)
  called as find(1,{delta,[{1,[{<<28,109,74,203,26,78,198,...>>,false}]}]})
in call from lasp_orset:remove_elem/3 (/Users/cmeiklejohn/Documents/lasp/_build/test/lib/lasp/src/lasp_orset.erl, line 359)
in call from lasp_lattice:lasp_orset_inflation_test/0 (/Users/cmeiklejohn/Documents/lasp/_build/test/lib/lasp/src/lasp_lattice.erl, line 615)
**error:function_clause
  output:<<"">>

      lasp_lattice: lasp_orset_strict_inflation_test...*failed*
in function orddict:find/2 (orddict.erl, line 86)
  called as find(1,{delta,[{1,[{<<49,81,71,121,135,224,49,...>>,false}]}]})
in call from lasp_orset:remove_elem/3 (/Users/cmeiklejohn/Documents/lasp/_build/test/lib/lasp/src/lasp_orset.erl, line 359)
in call from lasp_lattice:lasp_orset_strict_inflation_test/0 (/Users/cmeiklejohn/Documents/lasp/_build/test/lib/lasp/src/lasp_lattice.erl, line 635)
**error:function_clause
  output:<<"">>

      lasp_lattice: lasp_orset_gbtree_inflation_test...[0.003 s] ok
      lasp_lattice: lasp_orset_gbtree_strict_inflation_test...ok
      [done in 0.057 s]
    module 'lasp_leaderboard'
    module 'lasp_node_event_handler'
    module 'lasp_operations'
    lasp_orset: stat_test (module 'lasp_orset')...*failed*
in function orddict:find/2 (orddict.erl, line 86)
  called as find(<<"foo">>,{delta,[{<<"bar">>,[{<<142,225,74,60,14,43,86,...>>,false}]}]})
in call from lasp_orset:remove_elem/3 (/Users/cmeiklejohn/Documents/lasp/_build/test/lib/lasp/src/lasp_orset.erl, line 359)
in call from lasp_orset:stat_test/0 (/Users/cmeiklejohn/Documents/lasp/_build/test/lib/lasp/src/lasp_orset.erl, line 386)
**error:function_clause
  output:<<"">>
```

apps/lasp/src/lasp_orset.erl
```diff
@@ -304,6 +335,45 @@ minimum_tokens(Tokens) ->
...
+            end, orddict:new(), Elems),
+    {ok, {delta, OD}};
+update_delta({remove, Elem}, _Actor, ORDict) ->
+    remove_elem(Elem, ORDict, orddict:new());
+update_delta({remove_all, Elems}, _Actor, ORDict0) ->
+    remove_elems(Elems, ORDict0, orddict:new());
+update_delta({update, Ops}, Actor, ORDict) ->
+    apply_ops(Ops, Actor, ORDict).
+
+remove_elem(Elem, ORDict, DeltaORDict) ->
```
cmeiklejohn: This function was already in the file, but here it appears you are adding it again.  I think that is unnecessary and should be removed.

apps/lasp/src/lasp_orset.erl
```diff
@@ -304,6 +335,45 @@ minimum_tokens(Tokens) ->
...
+        {ok, Tokens} ->
+            Tokens1 = orddict:fold(fun(Token, _, Tokens0) ->
+                                           orddict:store(Token, true, Tokens0)
+                                   end, orddict:new(), Tokens),
+            {ok, {delta, orddict:store(Elem, Tokens1, DeltaORDict)}};
+        error ->
+            {error, {precondition, {not_present, Elem}}}
+    end.
+
+remove_elems([], _ORDict, DeltaORDict) ->
```
cmeiklejohn: This function was already in the file as well, and appears to have been added again.

apps/lasp/src/lasp_core.erl
```diff
@@ -337,6 +337,26 @@ bind(Id, Value, Store) ->
...
+    Mutator = fun(#dv{type=Type, metadata=Metadata0, value=Value0,
+                      waiting_threads=WT}=Object) ->
+            Metadata = MetadataFun(Metadata0),
+            %% Merge may throw for invalid types.
+            try
+                Merged = lasp_type:merge(Type, Value0, Value),
+                {ok, SW} = reply_to_all(WT, [], {ok, {Id, Type, Metadata, Merged}}),
+                NewObject = #dv{type=Type, metadata=Metadata, value=Merged, waiting_threads=SW},
+                %% Return value is a delta state
+                {NewObject, {ok, {Id, Type, Metadata, Value}}}
```
junghunyoo: bind() with {delta, Value} returns the delta state (Value) not the entire state (Merged). update() for the broadcast backend is using the state twice (one for the broadcast and the other for the return value). For the delta-mutation, the broadcast should use the delta state, but I am not sure about the return value of update(). Which state is the right one for the return value of update() with the delta-mutation (the delta state or the entire state)?

apps/lasp/src/lasp_orset.erl
junghunyoo: This function is different from the existing one. It has one extra parameter which is for storing delta states when the 'remove_all' operation is used.

apps/lasp/src/lasp_orset.erl
junghunyoo: Same as the remove_elem()

apps/lasp/src/lasp_core.erl
junghunyoo: bind() with {delta, Value} returns the delta state (Value) not the entire state (Merged) (see line 350). update() for the broadcast backend is using the state twice (one for the broadcast and the other for the return value). For the delta-mutation, the broadcast should use the delta state, but I am not sure about the return value of update(). Which state is the right one for the return value of update() with the delta-mutation (the delta state or the entire state)?

cmeiklejohn:
These changes are good; the failures that can be observed are related to a problem where Plumtree has a race on membership management on startup.

PR #72, opened 2016-02-25
Change the advertisement counter to use the delta operation for the g…

junghunyoo:
…-counter


cmeiklejohn:
@junghunyoo Do you mind making this use the `delta_mode` setting so we can use this test for both experiments?

apps/lasp/src/lasp_advertisement_counter.erl
```diff
@@ -209,31 +210,41 @@ server({#ad{counter=Counter}=Ad, _}, Ads) ->
...
+            client(SetType, CounterType, SyncInterval, Runner, Id, AdsWithContractsId, AdsWithContracts0, Counters, CountersDelta)
     after
         SyncInterval ->
-            {ok, AdsWithContracts, Counters} = synchronize(SetType, AdsWithContractsId, AdsWithContracts0, Counters0),
-            client(SetType, CounterType, SyncInterval, Runner, Id, AdsWithContractsId, AdsWithContracts, Counters)
+            %% Update dictionary.
+            Counters1 = dict:merge(fun(_Ad, OldCounter, CounterDelta) ->
+                                           CounterType:merge(OldCounter, CounterDelta)
+                                   end, Counters0, CountersDelta0),
+            {ok, AdsWithContracts, Counters} = synchronize(SetType, AdsWithContractsId, AdsWithContracts0, Counters1),
```
cmeiklejohn: This needs to be turned into a conditional so that based on `delta_mode` we can do delta or non-delta synchronization.  Otherwise, the changes look good.

cmeiklejohn:
@junghunyoo Upon further investigation, I determined that while you accumulate the Delta's here during the execution and periodically synchronize them, you don't actually transmit only the deltas, which causes this execution to use the same amount of bandwidth as the previous one.

cmeiklejohn:
Manually merged into the unstable branch so I can keep working towards the evaluation.

PR #81, opened 2016-03-08
Anti-entropy algorithm for the causal consistency of delta-CRDT

junghunyoo:


apps/lasp/src/lasp_core.erl
```diff
@@ -339,15 +346,31 @@ bind(Id, Value, Store) ->
...
+                      delta_map=DeltaMap0, delta_ack_map=AckMap}=Object) ->
             Metadata = MetadataFun(Metadata0),
             %% Merge may throw for invalid types.
             try
                 Merged = lasp_type:merge(Type, Value0, Value),
                 {ok, SW} = reply_to_all(WT, [], {ok, {Id, Type, Metadata, Merged}}),
-                NewObject = #dv{type=Type, metadata=Metadata, value=Merged, waiting_threads=SW},
-                %% Return value is a delta state
-                {NewObject, {ok, {Id, Type, Metadata, {delta, Value, Merged}}}}
+                OldValue = lasp_type:query(Type, Value0),
```
cmeiklejohn: How is this safe to propagate?  `query` returns the actual value and not the CRDT structure.  For a G-Counter, this is going to be just an integer; same for a PN-counter, so `NewValue` could equal `OldValue`, but they could be the results of completely different options, no?

junghunyoo: On Tue, 08 Mar 2016 14:32:55 -0800 "Christopher S. Meiklejohn"
<notifications@github.com> wrote

> >              Metadata = MetadataFun(Metadata0),
> >              %% Merge may throw for invalid types.
> >              try
> >                  Merged = lasp_type:merge(Type, Value0, Value),
> >                  {ok, SW} = reply_to_all(WT, [], {ok, {Id, Type, Metadata,
> >                  Merged}}), > -                NewObject = #dv{type=Type,
> >                  metadata=Metadata, value=Merged, waiting_threads=SW}, > - 
> >                                %% Return value is a delta state > -        
> >                                       {NewObject, {ok, {Id, Type, Metadata,
{delta, Value, Merged}}}}
> > +                OldValue = lasp_type:query(Type, Value0),
> 
> How is this safe to propagate?  'query' returns the actual value and not the
> CRDT structure.  For a G-Counter, this is going to be just an integer; same
> for a PN-counter, so 'NewValue' could equal 'OldValue', but they could be the
> results of completely different options, no? 

I was looking for the way of checking whether deltas are already merged, and I
think that if the old value (X) and the merged value (X join d) are equal, it
means that the deltas were merged to the old value before. I thought that I
should compare with actual values not crdt structures. If there is a better way
to compare them, I will use it. What do you think?

cmeiklejohn: If you compare with actual values, you will get false positives.  If the delta for a PN-counter for instance was a single increment and decrement, the value comparison would show that it was falsely already merged.

junghunyoo: I see. I will do the comparison based on the CRDT structure.

cmeiklejohn: Just use the ```lasp_lattice:is_strict_inflation``` call on the merged with delta and unmerged with delta value to determine if it's changed.

junghunyoo: Oh, I see. Actually, I tried with the lasp_lattice:is_inflation, but it was not working properly. I will ues the lasp_lattice:is_strict_inflation. Thanks.

cmeiklejohn: Yeah, inflation will return true for no change, but the strict inflation will only return true if it's actually different (and, monotonically greater.)

apps/lasp/src/lasp_core.erl
```diff
@@ -339,15 +346,27 @@ bind(Id, Value, Store) ->
...
+                        DeltaMap = orddict:store(Counter0, Value, DeltaMap0),
+                        NewObject = #dv{type=Type, metadata=Metadata, value=Merged,
+                                        waiting_threads=SW,
+                                        delta_counter=increment_counter(Counter0),
+                                        delta_map=DeltaMap, delta_ack_map=AckMap},
+                        %% Return value is a delta state.
+                        {NewObject, {ok, {Id, Type, Metadata, {delta, Value, Merged}}}};
+                    false ->
+                        %% Given delta state is already merged, no delta info update.
+                        {Object#dv{metadata=Metadata, value=Merged, waiting_threads=SW},
```
cmeiklejohn: If it's already been merged, why do you have to change the value to Merged here?  Can't you just omit the `value=Merged` section?  Do you need to change the metadata?

apps/lasp/src/lasp_core.erl
```diff
@@ -339,15 +346,27 @@ bind(Id, Value, Store) ->
...
             %% Merge may throw for invalid types.
             try
                 Merged = lasp_type:merge(Type, Value0, Value),
                 {ok, SW} = reply_to_all(WT, [], {ok, {Id, Type, Metadata, Merged}}),
-                NewObject = #dv{type=Type, metadata=Metadata, value=Merged, waiting_threads=SW},
-                %% Return value is a delta state
-                {NewObject, {ok, {Id, Type, Metadata, {delta, Value, Merged}}}}
+                case lasp_lattice:is_strict_inflation(Type, Value0, Merged) of
+                    true ->
+                        DeltaMap = orddict:store(Counter0, Value, DeltaMap0),
```
cmeiklejohn: Does this counter grow forever, or does it eventually get reset at some point?

apps/lasp/src/lasp_core.erl
```diff
@@ -830,6 +854,31 @@ reply_to_all([From|T], StillWaiting, Result) ->
 reply_to_all([], StillWaiting, _Result) ->
     {ok, StillWaiting}.
 
+receive_delta(Store, {delta_send, {Id, Type, _Metadata, Deltas},
```
cmeiklejohn: Can you add a `@doc` string about what this function does and when it's used?

apps/lasp/src/lasp_core.erl
```diff
@@ -830,6 +854,31 @@ reply_to_all([From|T], StillWaiting, Result) ->
...
+                      MetadataFunBind, MetadataFunDeclare}) ->
+    case do(get, [Store, Id]) of
+        {ok, _Object} ->
+            {ok, _Result} = bind(Id, Deltas, MetadataFunBind, Store);
+        {error, not_found} ->
+            {ok, _Result} = declare(Id, Type, MetadataFunDeclare, Store)
+    end,
+    ok;
+
+receive_delta(Store, {delta_ack, Id, From, Counter}) ->
```
cmeiklejohn: Can you add a `@doc` string about what this function does and when it's used?

apps/lasp/src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -233,25 +240,20 @@ update(Id, Operation, Actor) ->
...
-                             {delta, DeltaState};
-                         _ ->
-                             Value
-                     end,
     ReturnState = case Value of
                       {delta, _DeltaState, MergedState} ->
+                          %% No broadcasting for the delta.
                           MergedState;
                       _ ->
+                          case orddict:find(dynamic, Metadata) of
```
cmeiklejohn: I think these are nested wrong.  Don't you want to check if it's dynamic first?  Even if it's dynamic, you still want to avoid sending the deltas, too.

cmeiklejohn:
This looks good.

One concern though is that it appears from a quick reading of this that the set of deltas grows forever: we never have a way to purge the list of deltas that are waiting to be sent to clients.  This is going to be a problem if a client is partitioned for an extended period of time, correct?

If so, maybe we should do the following:

* Make it so that the set has a fixed size, and only stores some fixed number of deltas.  This option should be configurable.
* When clients synchronize, if they are too old, or have been disconnected for too long, transmit the full state instead.

Does that sound reasonable?

apps/lasp/src/lasp_core.erl
junghunyoo: I agree with your comment. I will remove the changes for the value and the metadata.

apps/lasp/src/lasp_core.erl
junghunyoo: As I know, this counter needs to grow forever.

apps/lasp/src/lasp_core.erl
junghunyoo: I will add it.

apps/lasp/src/lasp_core.erl
junghunyoo: I will add it.

vitorenesduarte:
You should be able to garbage collect deltas from time to time as long as you are exchanging acks.

If in the buffer you have the deltas resulting from updates 0-10 and ack map is something like
```json
acks["replica0"] = 5
acks["replica1"] = 10
acks["replica2"] = 8
```
you can delete the deltas 0-4. We still need the 5 because ```acks["replica0"] = 5```  does not mean it has seen the 5, only the 0-4.

I'm assuming this implementation follows the idea presented in the paper.

apps/lasp/src/lasp_plumtree_broadcast_distribution_backend.erl
junghunyoo: I will fix this.

junghunyoo:
The implementation for the garbage collection was intentionally left out in order to check the basic functionality first. The next PR will have it. Thanks.

cmeiklejohn:
@vitorenesduarte Agreed; however, I think my previous point still stands: if a client is disconnected for a long period of time, this list has the potential to grow quite large.  If that client never comes back, we've got a problem.  Therefore, I think we still need to bound the maximum number of deltas we can keep at any one point and fall back to state synchronization afterwards.

cmeiklejohn:
I think this merged, but it's unclear because GitHub is all wonky.

cmeiklejohn:
Manually merged.

vitorenesduarte:
@cmeiklejohn well thought! If you want to avoid sending the whole state to all clients just because one client disconnected, you can delete its entry from the ack map (this assures the buffer will not grow forever) and only for that client the whole state will be sent.
But this means clients should be able to detect those situations (other clients disconnected). Don't know if that's a possibility here.

PR #91, opened 2016-03-20
Garbage collection for the delta-based anti-entropy strategy

junghunyoo:
1. The delta_map has the limitation of the number of slots for storing deltas (`delta_mode_max_slots`).

2. When the slots are full, the first slot becomes an accumulator to store multiple deltas instead of deleting one of deltas (in order to avoid sending the entire state due to deleted deltas).

3. The delta_ack_map stores an extra boolean to check whether an 'ack' existed when the previous gc happened. With this boolean, we can check nodes are disconnected or slow and those nodes will be deleted during the next gc. This allows the deltas (garbage) to be collected after the disconnection of nodes.

4. The gc will happen after the certain number (`delta_mode_max_gc_counter`) of times of exchanges.

### DIFF
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -914,7 +914,7 @@ increment_counter(Counter) ->
 
 %% @private
 store_delta(Type, Counter, Delta, DeltaMap0) ->
-    MaxDeltaSlots = application:get_env(lasp, delta_mode_max_slots, 10),
+    MaxDeltaSlots = mochiglobal:get(delta_mode_max_slots, 10),
     %% Check the space of the DeltaMap
     case orddict:size(DeltaMap0) < MaxDeltaSlots of
         true ->
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -356,7 +356,7 @@ bind(Id, {delta, Value}, MetadataFun, Store) ->
                 {ok, SW} = reply_to_all(WT, [], {ok, {Id, Type, Metadata, Merged}}),
                 case lasp_lattice:is_strict_inflation(Type, Value0, Merged) of
                     true ->
-                        DeltaMap = orddict:store(Counter0, Value, DeltaMap0),
+                        DeltaMap = store_delta(Type, Counter0, Value, DeltaMap0),
                         NewObject = #dv{type=Type, metadata=Metadata, value=Merged,
                                         waiting_threads=SW,
                                         delta_counter=increment_counter(Counter0),
@@ -877,12 +877,12 @@ receive_delta(Store, {delta_ack, Id, From, Counter}) ->
     case do(get, [Store, Id]) of
         {ok, #dv{delta_ack_map=AckMap0}=Object} ->
             OldAck = case orddict:find(From, AckMap0) of
-                         {ok, Ack0} ->
+                         {ok, {Ack0, _GCed}} ->
                              Ack0;
                          error ->
                              0
                      end,
-            AckMap = orddict:store(From, max(OldAck, Counter), AckMap0),
+            AckMap = orddict:store(From, {max(OldAck, Counter), false}, AckMap0),
             do(put, [Store, Id, Object#dv{delta_ack_map=AckMap}]);
         _ ->
             error
@@ -911,6 +911,34 @@ write(Type, Value, Key, Store) ->
 %% @private
 increment_counter(Counter) ->
     Counter + 1.
+
+%% @private
+store_delta(Type, Counter, Delta, DeltaMap0) ->
+    MaxDeltaSlots = application:get_env(lasp, delta_mode_max_slots, 10),
+    %% Check the space of the DeltaMap
+    case orddict:size(DeltaMap0) < MaxDeltaSlots of
+        true ->
+            %% Store a new delta.
+            orddict:store(Counter, Delta, DeltaMap0);
+        false ->
+            DeltaCounterList = orddict:fetch_keys(DeltaMap0),
+            %% Find the minimum counter & its delta.
+            MinCounter0 = lists:nth(1, DeltaCounterList),
+            MinCounterDelta0 = orddict:fetch(MinCounter0, DeltaMap0),
+            %% Find the 2nd minimum counter & its delta.
+            MinCounter1 = lists:nth(2, DeltaCounterList),
+            MinCounterDelta1 = orddict:fetch(MinCounter1, DeltaMap0),
+            %% Merge them.
+            Merged = lasp_type:merge(Type,
+                                     MinCounterDelta0,
+                                     MinCounterDelta1),
+            %% Store the merged delta (minimum + 2nd minimum).
+            DeltaMap1 = orddict:store(MinCounter0, Merged, DeltaMap0),
+            %% Remove the 2nd minimum delta.
+            DeltaMap2 = orddict:erase(MinCounter1, DeltaMap1),
+            %% Store a new delta.
+            orddict:store(Counter, Delta, DeltaMap2)
+    end.
 
 -ifdef(TEST).
 
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -921,13 +921,9 @@ store_delta(Type, Counter, Delta, DeltaMap0) ->
             %% Store a new delta.
             orddict:store(Counter, Delta, DeltaMap0);
         false ->
-            DeltaCounterList = orddict:fetch_keys(DeltaMap0),
-            %% Find the minimum counter & its delta.
-            MinCounter0 = lists:nth(1, DeltaCounterList),
-            MinCounterDelta0 = orddict:fetch(MinCounter0, DeltaMap0),
-            %% Find the 2nd minimum counter & its delta.
-            MinCounter1 = lists:nth(2, DeltaCounterList),
-            MinCounterDelta1 = orddict:fetch(MinCounter1, DeltaMap0),
+            %% Find the minimum and 2nd minimum counters & those deltas.
+            [{MinCounter0, MinCounterDelta0}, {MinCounter1, MinCounterDelta1} | _Rest] =
+                orddict:to_list(DeltaMap0),
             %% Merge them.
             Merged = lasp_type:merge(Type,
                                      MinCounterDelta0,
--- a/src/lasp_plumtree_broadcast_distribution_backend.erl
+++ b/src/lasp_plumtree_broadcast_distribution_backend.erl
@@ -742,8 +742,6 @@ handle_cast({delta_send, From, {Id, Type, _Metadata, Deltas}, Counter},
                                           {Id, Type, _Metadata, Deltas},
                                           ?CLOCK_INCR,
                                           ?CLOCK_INIT}),
-    lager:info("Send Delta({delta_ack}): To: ~p, Counter: ~p",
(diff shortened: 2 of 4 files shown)


cmeiklejohn:
You need to rebase / merge with unstable, because some of the files have moved around.

junghunyoo:
I did rebase, but I am not sure I did it correctly. Could you check this?

cmeiklejohn:
Did you do it today?  Your branch has merge conflicts with `unstable`.

junghunyoo:
I did it 5 mins ago and without any conflicts. Maybe I did something wrong?

cmeiklejohn:
You didn't have the latest unstable, it seems, as you're missing those commits here.  I think you need to switch back to unstable, pull, then go back to this branch and rebase/merge.

src/lasp_core.erl
```diff
@@ -912,6 +912,34 @@ write(Type, Value, Key, Store) ->
 increment_counter(Counter) ->
     Counter + 1.
 
+%% @private
+store_delta(Type, Counter, Delta, DeltaMap0) ->
+    MaxDeltaSlots = application:get_env(lasp, delta_mode_max_slots, 10),
```
cmeiklejohn: This `gen_env` is extremely expensive.  Can you use mochiglobal like I did for the `delta_mode` setting?

src/lasp_core.erl
```diff
@@ -912,6 +912,34 @@ write(Type, Value, Key, Store) ->
...
+            %% Store a new delta.
+            orddict:store(Counter, Delta, DeltaMap0);
+        false ->
+            DeltaCounterList = orddict:fetch_keys(DeltaMap0),
+            %% Find the minimum counter & its delta.
+            MinCounter0 = lists:nth(1, DeltaCounterList),
+            MinCounterDelta0 = orddict:fetch(MinCounter0, DeltaMap0),
+            %% Find the 2nd minimum counter & its delta.
+            MinCounter1 = lists:nth(2, DeltaCounterList),
+            MinCounterDelta1 = orddict:fetch(MinCounter1, DeltaMap0),
```
cmeiklejohn: This section here, from L924-930 is traversing the list multiple times and is very expensive.  (Once to grab all keys, again to get the first element, then again to get it from the dict, then again to get the second element, then again to get the delta from the dict.)  Can we just convert it to a list once with `to_list` at the beginning of the execution and operate directly on it as a list?  

When you want to grab the first two minimal counters, you could easily do [First, Second | _Rest] and save a bit of computation that you have in your example above.

src/lasp_core.erl
```diff
@@ -912,6 +912,34 @@ write(Type, Value, Key, Store) ->
...
+            %% Merge them.
+            Merged = lasp_type:merge(Type,
+                                     MinCounterDelta0,
+                                     MinCounterDelta1),
+            %% Store the merged delta (minimum + 2nd minimum).
+            DeltaMap1 = orddict:store(MinCounter0, Merged, DeltaMap0),
+            %% Remove the 2nd minimum delta.
+            DeltaMap2 = orddict:erase(MinCounter1, DeltaMap1),
+            %% Store a new delta.
+            orddict:store(Counter, Delta, DeltaMap2)
```
cmeiklejohn: So, it appears the general idea here is that every single time an update is performed, the earliest delta is collapsed in to the next one to open a slot for the newest update, is that correct?

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -187,7 +188,14 @@ exchange(Peer) ->
     case mochiglobal:get(delta_mode, false) of
         true ->
             %% Anti-entropy mechanism for causal consistency of delta-CRDT.
-            gen_server:call(?MODULE, {exchange, Peer}, infinity);
+            {ok, Pid, GC_Counter} = gen_server:call(?MODULE, {exchange, Peer}, infinity),
+            MaxGCCounter = mochiglobal:get(delta_mode_max_gc_counter, 7),
```
cmeiklejohn: This default of 7 should be in an include file and documented regarding to how it works and what it's for.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -187,7 +188,14 @@ exchange(Peer) ->
...
+            case GC_Counter == MaxGCCounter of
```
cmeiklejohn: Erlang, more stylistically, would have this variable be `GCCounter` and not `GC_Counter`; no underscores.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -651,6 +658,78 @@ handle_call(exchange, _From, #state{store=Store}=State) ->
...
+handle_call(delta_gc, _From, #state{store=Store}=State) ->
+    Function =
+        fun({Id, #dv{delta_map=DeltaMap0, delta_ack_map=AckMap0,
+                     delta_counter=Counter}=_Object},
+            Acc0) ->
+                case orddict:is_empty(DeltaMap0) of
+                    true ->
+                        Acc0;
+                    false ->
+                        %% Remove disconnected or slow node() entries from the AckMap.
```
cmeiklejohn: Once they are removed from the `AckMap`, are they added back again?  When?

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -651,6 +658,78 @@ handle_call(exchange, _From, #state{store=Store}=State) ->
...
+                        %% disconnected or slow: seen the previous GC & no change
+                        RemovedAckMap0 =
+                            orddict:filter(fun(_Node, {Ack, GCed}) ->
+                                                   (GCed == false) or (Ack == Counter)
```
cmeiklejohn: `or` needs to be `orelse`: Erlang-ism.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -651,6 +658,78 @@ handle_call(exchange, _From, #state{store=Store}=State) ->
...
+                                    lists:min([Ack || {_Node, {Ack, _GCed}} <- RemovedAckMap]),
+                                %% size() should be bigger than 0.
+                                MinAck1 =
+                                    case orddict:size(DeltaMap0) of
+                                        1 ->
+                                            MinAck;
+                                        _ ->
+                                            Counters = orddict:fetch_keys(DeltaMap0),
+                                            NotCollectable =
+                                                (MinAck > lists:nth(1, Counters)) and
```
cmeiklejohn: `and` needs to be `andalso`

src/lasp_core.erl
```diff
@@ -914,7 +914,7 @@ increment_counter(Counter) ->
 
 %% @private
 store_delta(Type, Counter, Delta, DeltaMap0) ->
-    MaxDeltaSlots = application:get_env(lasp, delta_mode_max_slots, 10),
+    MaxDeltaSlots = mochiglobal:get(delta_mode_max_slots, 10),
```
cmeiklejohn: You also need to set it in the `lasp_sup` file, else it will always be undefined.

src/lasp_core.erl
junghunyoo: Not for every update, only when there is no space in the delta_map.

cmeiklejohn: Yes, right.  I understand.

src/lasp_plumtree_broadcast_distribution_backend.erl
junghunyoo: When the 'ack' is delivered again from those nodes, they are added again.

cmeiklejohn: After they've been removed and they attempt to sync, are they sent the entire state?

junghunyoo: Yes. If the node is not in the ack_map, it always sends the entire state.

cmeiklejohn: Great, sounds fantastic.

src/lasp_core.erl
```diff
@@ -912,6 +912,30 @@ write(Type, Value, Key, Store) ->
...
+            [{MinCounter0, MinCounterDelta0}, {MinCounter1, MinCounterDelta1} | _Rest] =
+                orddict:to_list(DeltaMap0),
+            %% Merge them.
+            Merged = lasp_type:merge(Type,
+                                     MinCounterDelta0,
+                                     MinCounterDelta1),
+            %% Store the merged delta (minimum + 2nd minimum).
+            DeltaMap1 = orddict:store(MinCounter0, Merged, DeltaMap0),
+            %% Remove the 2nd minimum delta.
+            DeltaMap2 = orddict:erase(MinCounter1, DeltaMap1),
```
cmeiklejohn: I'm confused.  Doesn't this leave a gap?  For instance, if the bottom two elements in the orddict were 2, 3, it seems like you delete element 3 and update 2.  Wouldn't it be better to do the opposite to keep the numbering without gaps?  For instance, delete 2 after you merge 2 into 3?

src/lasp_sup.erl
```diff
@@ -147,6 +147,12 @@ init(_Args) ->
                             lasp_plumtree_broadcast_distribution_backend),
     mochiglobal:put(distribution_backend, DistributionBackend),
 
+    MaxDeltaSlots = application:get_env(?APP, delta_mode_max_slots, 10),
+    mochiglobal:put(delta_mode_max_slots, MaxDeltaSlots),
+
+    MaxGCCounter = application:get_env(?APP, delta_mode_max_gc_counter, 7),
```
cmeiklejohn: You could probably use `MAX_GC_COUNTER` here instead of the literal 7.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -651,6 +659,78 @@ handle_call(exchange, _From, #state{store=Store}=State) ->
...
+                                %% size() should be bigger than 0.
+                                MinAck1 =
+                                    case orddict:size(DeltaMap0) of
+                                        1 ->
+                                            MinAck;
+                                        _ ->
+                                            Counters = orddict:fetch_keys(DeltaMap0),
+                                            NotCollectable =
+                                                (MinAck > lists:nth(1, Counters)) andalso
+                                                    (MinAck < lists:nth(2, Counters)),
```
cmeiklejohn: It seems like the case where these are equal is not considered.  Should it be?

cmeiklejohn:
Some more minor nitpicks, but other than that this strategy looks good!

src/lasp_core.erl
junghunyoo: The reason why the accumulator has a smaller (start) counter is that a larger (end) counter can be retrieved by the next counter (next counter - 1). In order to use the accumulator slot properly, we need both counters (start & end).

src/lasp_plumtree_broadcast_distribution_backend.erl
junghunyoo: This comparison is needed when the `MinAck` is in the accumulator slot (then, the `MinAck` will be adjusted). If it is equal to boundaries, the rest of code for garbage collection will handle it properly.

PR #94, opened 2016-04-04
'not_found' error handling for update(), bind() and read()

junghunyoo:


src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -261,24 +261,35 @@ query(Id) ->
...
+                                      _ ->
+                                          broadcast({Id, Type, Metadata, Value}),
+                                          Value
+                                  end
+                          end,
+            {ok, {Id, Type, Metadata, ReturnState}};
+        {error, not_found} ->
+            case Id of
+                {StorageId, TypeId} ->
+                    declare(StorageId, TypeId),
```
cmeiklejohn: I think you should probably assert the return of this call.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -261,24 +261,35 @@ query(Id) ->
...
+                    gen_server:call(?MODULE, {update, Id, Operation, Actor}, infinity);
```
cmeiklejohn: Can't the update itself do the declare, so we don't need to call declare and then update?

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -325,7 +347,19 @@ bind_to(Id, TheirId) ->
 %%
 -spec read(id(), threshold()) -> {ok, var()} | error().
 read(Id, Threshold) ->
-    gen_server:call(?MODULE, {read, Id, Threshold}, infinity).
+    case gen_server:call(?MODULE, {read, Id, Threshold}, infinity) of
+        {error, not_found} ->
+            case Id of
+                {StorageId, TypeId} ->
+                    declare(StorageId, TypeId),
+                    gen_server:call(?MODULE, {read, Id, Threshold}, infinity);
```
cmeiklejohn: Similar comments above about the two step process -- can we just make the update implicitly call the declare to make this more streamlined?

src/lasp_plumtree_broadcast_distribution_backend.erl
junghunyoo: I will add it.

src/lasp_plumtree_broadcast_distribution_backend.erl
junghunyoo: Before I made this change, I tried to add the `declare()` & `read()` to the return value of `do(update...)` which is inside of the `Mutator()` func in `lasp_core:read()`. However, calling the `read()` again was not successful. Did the update itself mean after checking the return value of `do(update...)`, right?

cmeiklejohn: Why was it unsuccessful?

junghunyoo: At the end of the minimal_delta_test test, I saw this error:

2016-04-05 13:42:19.628 [error] <18741.218.0> CRASH REPORT Process <18741.218.0> with 0 neighbours exited with reason: no such process or port in call to gen_server:call(lasp_plumtree_broadcast_distribution_backend, {bind,{<<250,81,209,51,216,79,178,115,220,176,248,59,51,124,157,100,220,101,51,11>>,lasp_gcounter},...}, infinity) in gen_server:terminate/7 line 826

cmeiklejohn: That appears to be a race condition on shutting down -- something is calling the bind function after the lasp_plumtree_broadcast_distribution_backend process has terminated.

junghunyoo: If you still want to handle the declare & re-read inside of the read(), I will do more investigation.

cmeiklejohn: Why do you have to re-read?

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -861,6 +899,20 @@ collect_deltas(Type, DeltaMap, Min0, Max) ->
                           end, lasp_type:new(Type), SmallDeltaMap),
     {delta, Deltas}.
 
+%% @private
+not_found_handling({error, not_found}, Id, #state{store=Store, actor=Actor}=_State,
```
cmeiklejohn: If `_State` is ignored, you could just remove it.

junghunyoo: I see. I will remove it.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -861,6 +899,20 @@ collect_deltas(Type, DeltaMap, Min0, Max) ->
                           end, lasp_type:new(Type), SmallDeltaMap),
     {delta, Deltas}.
 
+%% @private
+not_found_handling({error, not_found}, Id, #state{store=Store, actor=Actor},
```
cmeiklejohn: Can we rename this to something like ```declare_if_not_found```?

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -516,9 +526,19 @@ handle_call({bind, Id, Value}, _From,
 %% this could change if the other module is later refactored.
 handle_call({bind, Id, Metadata0, Value}, _From,
             #state{store=Store, counter=Counter}=State) ->
-    {_Time, Result} = timer:tc(?CORE,
-                               bind,
-                               [Id, Value, ?CLOCK_MERG, Store]),
+    {_Time, Result0} = timer:tc(?CORE,
```
cmeiklejohn: You can remove the ```timer:tc``` call here.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -503,9 +503,19 @@ handle_call({query, Id}, _From, #state{store=Store}=State) ->
 %% broadcast the result.
 handle_call({bind, Id, Value}, _From,
             #state{store=Store, actor=Actor, counter=Counter}=State) ->
-    {_Time, Result} = timer:tc(?CORE,
-                               bind,
-                               [Id, Value, ?CLOCK_INCR, Store]),
+    {_Time, Result0} = timer:tc(?CORE,
```
cmeiklejohn: You can remove the ```timer:tc``` call here.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -861,6 +899,20 @@ collect_deltas(Type, DeltaMap, Min0, Max) ->
...
+%% @private
+not_found_handling({error, not_found}, Id, #state{store=Store, actor=Actor},
+                   Module, Function, Args) ->
+    case Id of
+        {StorageId, TypeId} ->
+            {ok, {Id, _, _, _}} = ?CORE:declare(StorageId, TypeId,
+                                                ?CLOCK_INIT, Store),
+            erlang:apply(Module, Function, Args);
+        _ ->
+            {error, not_found}
```
cmeiklejohn: Can you leave a comment as to why we would ever hit this case?

cmeiklejohn:
Looks good!

PR #97, opened 2016-04-25
Initial implementation of the incremental computation for map()

junghunyoo:
A new test SUITE for incremental computation testing was also added.

### DIFF
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -30,6 +30,7 @@
          bind_to/3,
          read/2,
          read/3,
+         read_delta/3,
          read_any/2,
          declare/1,
          declare/2,
@@ -189,6 +190,32 @@ read(Id, Threshold, Store) ->
                 end
             end,
     read(Id, Threshold, Store, Self, ReplyFun, BlockingFun).
+
+%% @doc Perform a delta read for a particular identifier.
+%%
+%%      Given an `Id', perform a blocking read until a delta is
+%%      delivered.
+%%
+%%      This operation blocks until any deltas have been delivered.
+%%
+-spec read_delta(id(), value(), store()) -> {ok, {delta, var()}}.
+read_delta(Id, Threshold, Store) ->
+    case Threshold of
+        {strict, undefined} ->
+            %% The first run uses the current value.
+            lasp:read(Id, undefined);
+        {strict, _} ->
+            %% From the second run, it always waits for a delta
+            %% (from the local or the remote).
+            Self = self(),
+            {ok, Object} = do(get, [Store, Id]),
+            WDT = lists:append(Object#dv.waiting_delta_threads, [{delta, Self}]),
+            do(put, [Store, Id, Object#dv{waiting_delta_threads=WDT}]),
+            receive
+                X ->
+                    X
+            end
+    end.
 
 %% @doc Perform a monotonic read for a series of given idenfitiers --
 %%      first response wins.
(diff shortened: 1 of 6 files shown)


src/lasp_core.erl
```diff
@@ -190,6 +191,32 @@ read(Id, Threshold, Store) ->
...
+            %% The first run uses the current value.
+            lasp:read(Id, undefined);
+        {strict, _} ->
+            %% From the second run, it always waits for a delta
+            %% (from the local or the remote).
+            Self = self(),
+            {ok, Object} = do(get, [Store, Id]),
+            WDT = lists:append(Object#dv.waiting_delta_threads, [{delta, Self}]),
+            do(put, [Store, Id, Object#dv{waiting_delta_threads=WDT}]),
+            receive
```
cmeiklejohn: Eventually, this should be parameterized as the other ```read``` operation is.  Maybe leave a ```@todo``` here so we remember?

src/lasp_core.erl
```diff
@@ -701,7 +732,14 @@ map(Id, Function, AccId, Store, BindFun, ReadFun) ->
                                  lasp_orset ->
                                      lasp_orset:map(Function, V)
                              end,
-                  {ok, _} = BindFun(AccId, AccValue, Store)
+                  {ok, _} = BindFun(AccId, AccValue, Store);
+             %% A delta from the input will generate a new delta for the output
```
cmeiklejohn: Can you clarify what this comment means?

cmeiklejohn:
Do you have any way to gauge the performance difference here?

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -387,7 +391,12 @@ intersection(Left, Right, Intersection) ->
 %%
 -spec map(id(), function(), id()) -> ok | error().
 map(Id, Function, AccId) ->
-    gen_server:call(?MODULE, {map, Id, Function, AccId}, infinity).
+    case lasp_config:get(delta_mode, false) of
```
cmeiklejohn: Maybe we control this using another variable as well, like to enable or disable incremental computation?  We need a way to evaluate deltas with this enabled and with this disabled.

src/lasp_core.erl
junghunyoo: When the `process()` receives a delta of the input from  the `ReadFun()`, the `lasp_orset:map()` transforms it to a delta of the output. And `bind()` will add the delta to the output. This is what I want to say in that comment. I will update it :)

src/lasp_plumtree_broadcast_distribution_backend.erl
junghunyoo: I will add a new variable for the incremental computation.

junghunyoo:
In the `lasp_delta_SUITE`, there are two test cases. The first one is testing ORSet without the `delta_mode` and the incremental computation. And the second one enables the `delta_mode` and the incremental computation. Both cases measure the elapsed time for the `read()` of the output after making some changes to the input.

This is my latest results:
Time without incremental computation: 1262547
Time with incremental computation: 657246

One more thing... In order to see the benefits of the incremental computation, the original input set should be large enough (my test cases have 10,000 elements before making some changes to the input).

cmeiklejohn:
That's total elapsed time, right?

It would be interesting to generate a graph of elapsed time for each operation -- each time an element is added the resulting transformation.  We *should* see based on that linear performance of the map O(n) and the improvement to O(1) when the incremental bits are enabled.

junghunyoo:
The result time is only for the `read()` of the output after making an update for the input. I used the `timer:tc()` for it. You can check it in the 'lasp_delta_SUITE' file. Once I make the change for a new variable of the incremental computation, I will try to generate a graph with various tests.

cmeiklejohn:
:+1:, good work!

PR #108, opened 2016-05-27
Dependency graph

ergl:
This pull request:

- Adds a prototype version of the dependency graph (5fb4b28afc52229aab73b0899c8b51fde0fe810d and b5454bf9654eca4a29933234e51aa15d6a2fe17d).

- Adds a "write" step to the lasp process (c8ccf43b5a63362e73e50b2d13472893aca8ecbe, 098a60389dc0e0ba18afcc80c703c771e2aead5e).
  This is what we discussed the other day.

    >Additional support for lasp processes in 3 steps: Read, Transform and
    >Write.

    >Now, instead of passing a single function that computes the changes and
    >then writes them to the appropiate CRDT, we pass two functions, one that
    >computes the new value, and another that writes the value.

- Tracks lasp processes as edges in the dag (1a4968944c12e0ff7cf26626cc5dc94c7ec4d0b4).
  It won't create the process if it detects that it will form a loop, but won't
  return an error.

  Additionally, it adds the input(s) and output vertices to the graph, but we
  should move this to declare and declare dynamic.

- Tracks `map`, `filter`, `stream`, `union`, `intersection`, `product` and `bind_to`.
  This is done by using the `lasp_process:start_dag_link/1` function, that
  expects a list of readfuns, a transforming function, and a write function.

  Todo: track `fold`.

### DIFF
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -737,10 +737,11 @@ stream(Id, Function, Store) ->
 %%      stream can result in observable nondeterminism.
 %%
 stream(Id, Function, _Store, ReadFun) ->
-    Fun = fun({_, T, _, V}) ->
-            Function(lasp_type:query(T, V))
+    TransFun = fun({_, T, _, V}) ->
+        Function(lasp_type:query(T, V))
     end,
-    lasp_process:start_link([[{Id, ReadFun}], Fun]).
+    WriteFun = fun(_, X) -> X end,
+    lasp_process:start_dag_link([{Id, ReadFun}], TransFun, {undefined, WriteFun}).
 
 %% @doc Callback wait_needed function for lasp_vnode, where we
 %%      change the reply and blocking replies.
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -305,7 +305,7 @@ query(Id, Store) ->
 %%
 -spec bind_to(id(), id(), store()) -> {ok, pid()}.
 bind_to(AccId, Id, Store) ->
-    bind_to(AccId, Id, Store, ?BIND, ?READ).
+    bind_to(AccId, Id, Store, ?WRITE, ?READ).
 
 %% @doc Spawn a function.
 %%
@@ -560,10 +560,8 @@ bind_to(AccId, Id, Store, BindFun) ->
     bind_to(AccId, Id, Store, BindFun, ?READ).
 
 bind_to(AccId, Id, Store, BindFun, ReadFun) ->
-    Fun = fun({_, _, _, V}) ->
-        {ok, _} = BindFun(AccId, V, Store)
-    end,
-    lasp_process:start_link([[{Id, ReadFun}], Fun]).
+    TransFun = fun({_, _, _, V}) -> V end,
+    lasp_process:start_dag_link([{Id, ReadFun}], TransFun, {AccId, BindFun(Store)}).
 
 %% @doc Fold values from one lattice into another.
 %%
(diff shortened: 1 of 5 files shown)


src/lasp_core.erl
```diff
@@ -77,6 +77,12 @@
                 ?MODULE:bind(_AccId, AccValue, _Store)
               end).
 
+-define(WRITE, fun(_Store) ->
```
cmeiklejohn: How is write different from bind?

src/lasp_core.erl
```diff
@@ -554,10 +560,8 @@ bind_to(AccId, Id, Store, BindFun) ->
     bind_to(AccId, Id, Store, BindFun, ?READ).
 
 bind_to(AccId, Id, Store, BindFun, ReadFun) ->
-    Fun = fun({_, _, _, V}) ->
-        {ok, _} = BindFun(AccId, V, Store)
-    end,
-    lasp_process:start_link([[{Id, ReadFun}], Fun]).
+    TransFun = fun({_, _, _, V}) -> V end,
```
cmeiklejohn: Can you explain the meaning behind this?  It looks like you're dropping all of the type information and the metadata...

src/lasp_core.erl
ergl: Write is only there to generate the bind function with the `Store` captured in a closure, so we don't have to pass that to the lasp process as part of the writing function

src/lasp_core.erl
ergl: I assumed it was safe to do so since the previous implementation only cared about the input value. 

Seems like `bind_to(S2, S1)` is the same as `map(S1, fun(X) -> X end, S2)`, so the transforming function just returns the input value, and the write function binds it to S2.

cmeiklejohn: Yes, `bind_to` originally had different behavior, but now it's similar to a map.  It isn't *precisely* map, because it works for all data types, where map only operates on collection types.

src/lasp_core.erl
cmeiklejohn: Does this fully replace `bind` or is `bind` still used other places?

src/lasp_dependence_dag.erl
```diff
@@ -0,0 +1,202 @@
...
+-spec add_vertices(list(id())) -> ok.
+add_vertices([]) ->
+    ok;
+
+add_vertices(Vs) ->
+    gen_server:call(?MODULE, {add_vertices, Vs}, infinity).
+
+%% @doc Check if linking the given vertices will form a loop.
+-spec is_loop(list(id()), id()) -> boolean().
+is_loop(Src, Dst) ->
```
cmeiklejohn: What is meant by loop here?

cmeiklejohn:
Looks 👍 overall.  Just need a clarification on some of the components; left some comments.

src/lasp_core.erl
ergl: It is still used by `fold`, I haven't tested that it works with `start_dag_link` yet.

src/lasp_dependence_dag.erl
ergl: I mixed up my terminology here, it detects edges connecting a vertex with itself, but it also detects cycles in the graph.

cmeiklejohn: Are there more mathematical/graph theory terms for what we're attempting to identify here?  Sounds like a strongly connected component, maybe?

ergl: That's correct, it will check that the result of linking any of the given vertices
won't introduce a strongly connected component in the graph.  Even though
a strongly connected graph doesn't include (as far as I know) a vertex with an
edge to itself, this will also check that.

Most of the work is done by the `digraph:add_edge/3` function, that returns
`{error, {bad_edge, Path}}` if an edge between the given vertices introduces
a cycle in the graph.

cmeiklejohn: OK, can we rename the loop terminology to be closer to this to be clearer in the code before we merge?

cmeiklejohn:
👍

PR #110, opened 2016-05-30
Monitor lasp processes

ergl:
Since lasp processes might terminate (as they are marked transient) or
get restarted, we have to be able to update the labels in the edges that
represent those processes.

The `lasp_dependence_dag` module monitors all pids that get added as
labels in an edge of the graph. When one of those monitored processes
dies, for whatever reason, we remove all the edges that reference that
pid.

Those processes that get restarted will join the dag again as soon as
they execute their `init` function.


src/lasp_core.erl
```diff
@@ -269,6 +269,8 @@ declare(Id, Type, MetadataFun, MetadataNew, Store) ->
             %% Do nothing; make declare idempotent at each replica.
             {ok, {{Id, Type}, Type, Metadata, Value}};
         _ ->
+            %% @todo is this the correct place?
```
cmeiklejohn: Yes, I believe this is correct.

However, you also probably should look at the `declare_if_not_found` function as well, which will declare a variable that isn't declared yet on read if the message happened to be delayed for some reason.

src/lasp_dependence_dag.erl
```diff
@@ -11,7 +11,9 @@
          add_vertices/1]).
 
 %% Test
+%% @todo Only export on test.
```
cmeiklejohn: Hrm?

src/lasp_core.erl
ergl: Seems like `declare_if_not_found` calls  `declare/4` in core, so this change should be enough, right?

src/lasp_dependence_dag.erl
ergl: These functions are useful to check that the dag is correct, but I don't think they should be part of the public api. I'd surround the export with an `ifdef(TEST)` once I set up some test suite for this module.

src/lasp_core.erl
cmeiklejohn: Yes.

cmeiklejohn:
Needs rebase before merging.

ergl:
I've added the config option to this PR since it needed a bunch of changes from this one.

Also, it would be good to get #112 merged before so I can also add the config check to the
`lasp_dag_resource` module as well.

PR #111, opened 2016-05-31
Add dot representation of the dag

ergl:
Add options to export the current state of the dependency graph as a graphviz dot file.

Safe to merge before #110


cmeiklejohn:
Do you think it would be possible to wire this up to the web interface that's running?

ergl:
Should be done now, there are for sure a couple of things that could be fixed in the html and javascript, though.

PR #113, opened 2016-06-02
Add config check to lasp_dag_resource

ergl:


src/lasp_dag_resource.erl
```diff
@@ -13,8 +13,11 @@ content_types_provided(Req, Ctx) ->
     {[{"application/json", to_json}], Req, Ctx}.
 
 to_json(ReqData, State) ->
-    Status = case lasp_dependence_dag:to_dot() of
-        {ok, Content} -> #{present => true, dot_content => Content};
-        _ -> #{present => false}
+    Status = case lasp_config:get(dag_enabled, false) of
```
cmeiklejohn: Are you sure you are doing a `lasp_config:set` somewhere in the code for this to be set?

ergl: I added it with the previous PR
https://github.com/lasp-lang/lasp/blob/unstable/src/lasp_sup.erl#L121

cmeiklejohn: Sounds good; just checking to make sure I didn't miss it.

PR #115, opened 2016-06-03
Add function metadata to the dag

ergl:
Adds the read, write and transform functions as metadata to the dag.

Since we now have metadata, my next step is to detect lasp processes
that perform the same work (same edges, same transforming function)
so that we don't count them as two separate childs.

After that, the next optimization will detect all lasp processes connecting
(directly) two nodes and merge them into an union of all of them.

### DIFF
--- a/src/lasp_dependence_dag.erl
+++ b/src/lasp_dependence_dag.erl
@@ -308,19 +308,19 @@ delete_with_pid(Graph, Src, Dst, Pid) ->
 %%      only the ones linking to V2.
 %%
 -spec get_direct_edges(digraph:graph(),
-    digraph:vertex(), digraph:vertex()) -> list(edge()).
+                       digraph:vertex(), digraph:vertex()) -> list(edge()).
 
 get_direct_edges(G, V1, V2) ->
-  case directly_connected(G, V1, V2) of
-    false -> [];
-    true ->
-      lists:flatmap(fun(Ed) ->
-        case digraph:edge(G, Ed) of
-          {_, _, To, _}=E when To =:= V2 -> [E];
-          _ -> []
-        end
-                    end, digraph:out_edges(G, V1))
-  end.
+    case directly_connected(G, V1, V2) of
+        false -> [];
+        true ->
+            lists:flatmap(fun(Ed) ->
+                case digraph:edge(G, Ed) of
+                    {_, _, To, _}=E when To =:= V2 -> [E];
+                    _ -> []
+                end
+            end, digraph:out_edges(G, V1))
+    end.
 
 %% @doc Are V1 and V2 linked directly?
 %%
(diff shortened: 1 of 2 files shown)


ergl:
Actually, I don't see how the extra information couldn't be encoded directly into
the labels of the edges. I'll change that.

ergl:
Alright, everything should be good now

src/lasp_dependence_dag.erl
```diff
@@ -82,9 +88,9 @@ will_form_cycle(Src, Dst) ->
 %%      either because it formed a loop, or because some of the
 %%      vertices weren't in the graph.
 %%
--spec add_edges(list(id()), id(), pid()) -> ok | error.
-add_edges(Src, Dst, Pid) ->
-    gen_server:call(?MODULE, {add_edges, Src, Dst, Pid}, infinity).
+-spec add_edges(list(id()), id(), pid(), list(term())) -> ok | error.
```
cmeiklejohn: Can you be more specific then `term()` here?

src/lasp_dependence_dag.erl
```diff
@@ -205,16 +211,31 @@ handle_call({will_form_cycle, From, To}, _From, #state{dag=Dag}=State) ->
 %%
 %%      We monitor all edge Pids to know when they die or get restarted.
 %%
-handle_call({add_edges, Src, Dst, Pid}, _From, #state{dag=Dag, process_map=Pm}=State) ->
-    %% @todo Add metadata and duplicate checking.
-    Status = [digraph:add_edge(Dag, V, Dst, Pid) || V <- Src],
+handle_call({add_edges, Src, Dst, Pid, [ReadFuns, TransFun, {Dst, WriteFun}]},
+            _From, #state{dag=Dag, process_map=Pm}=State) ->
+
+    %% @todo Add duplicate checking.
```
cmeiklejohn: Can you give me an example where a duplicate would occur?  Is it just if someone literally runs the same command multiple times?

ergl: Yes, that's exactly it. We don't want the dag to think a vertex has more childs than it truly does.

cmeiklejohn: Sure, but, if they run the command twice it is actually creating *two* processes, right?

ergl: Right. My idea was to kill the old process when a new one gets added to the graph.

cmeiklejohn: OK, that seems reasonable.

cmeiklejohn: Should this wait to be merged until you make that change?

ergl: I'll submit that in a new PR.

Just pushed a change for the `add_edges` type, but I think this can be merged already.

PR #116, opened 2016-06-05
Merge all emanating edges into a single one

ergl:
For all vertices in the dag, merge all emanating edges representing
unary functions into a single one, that performs the union of all
functions in the old edges. Non-reversible.

Incident edges on the child emanating from other nodes are preserved.

### DIFF
--- a/src/lasp_dependence_dag.erl
+++ b/src/lasp_dependence_dag.erl
@@ -346,9 +346,12 @@ collect_transform_functions(Edges) ->
 %%      into a new process that represents the union of all of them.
 %%
 -spec merge_unary(id(), id(), digraph:graph()) -> ok | pid().
-merge_unary(Src, Dst, G) ->
+merge_unary({_, orset}=Src, Dst, G) ->
     Edges = get_direct_edges_with_arity(G, Src, Dst, 1),
-    union_unary(Edges).
+    union_unary(Edges);
+
+merge_unary(_, _, _) ->
+    ok.
 
 %% @doc Given a list of edges, merge them together into a single one.
 %%
--- a/src/lasp_dependence_dag.erl
+++ b/src/lasp_dependence_dag.erl
@@ -214,7 +214,6 @@ handle_call({will_form_cycle, From, To}, _From, #state{dag=Dag}=State) ->
 handle_call({add_edges, Src, Dst, Pid, ReadFuns, TransFun, {Dst, WriteFun}},
             _From, #state{dag=Dag, process_map=Pm}=State) ->
 
-    %% @todo Add duplicate checking.
     %% For all V in Src, make edge (V, Dst) with label {Pid, Read, Trans, Write}
     %% (where {Id, Read} = ReadFuns s.t. Id = V)
     Status = lists:map(fun(V) ->
@@ -234,6 +233,9 @@ handle_call({add_edges, Src, Dst, Pid, ReadFuns, TransFun, {Dst, WriteFun}},
             ProcessMap = lists:foldl(fun(El, D) ->
                 dict:append(Pid, {El, Dst}, D)
             end, Pm, Src),
+
+            %% Merge all edges representing unary functions together
+            lists:foreach(fun(V) -> merge_unary(V, Dst, Dag) end, Src),
 
             {ok, State#state{process_map=ProcessMap}}
     end,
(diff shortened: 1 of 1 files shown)


ergl:
One core assumption of this change is that read and write functions
in all edges are the same, although I'm sure this could be changed
if it posed a problem.

ergl:
Regarding what I previously said on #115 about checking for "duplicate"
processes, it turns out that even though read, transform and write functions
might be the same, erlang won't believe they represent the same term.

Sometimes, the same function reference gave different values for `phash2`, or
the terms themselves changed (functions generated in the plumtree module
change their reference at random intervals).

As this approach unions all functions, the final result is the same.

cmeiklejohn:
Can you explain more context about what the goal is here? 

I'm still not clear *why* precisely you are combining all unary functions into a single function.  Can you give a little intuition behind the change and why it's valuable?

ergl:
Sure. Our original goal is to remove intermediate vertices from the graph. One way
to see if a given vertex is safe to remove is by counting the number of children it has
(how many vertices are currently reading from it).

However, to count how many children a vertex has in our current implementation of
the graph, one has to get the list of all emanating edges, and then count how many
distinct vertices we reach.

The goal of this change is to guarantee that the outdegree of a vertex is exactly the
number of children it has. For that to hold, binary and unary functions have to merged
together as well, too.

However, merging unary functions may be enough. With this change, we guarantee
that if a given vertex has an outdegree greater than 1, it means:

1. it has more than one child (it can't be removed), or

2. it has one child with more than one parent.

Removing vertices with more than one parent can be done too, but I figured a good
first approach was to focus only on those with just one parent.

ergl:
While looking for solutions for our original problem, I found [edge contraction](https://en.wikipedia.org/wiki/Edge_contraction).

This process can introduce multiple edges even if the original graph is simple (no
multiple edges between the same pair of vertices), but by removing the possibility
of creating multiple edges, we guarantee that the resulting graph is simple too.

Given that our optimizations are dynamic in nature, we might want to optimize
and already optimized graph when a suitable set of vertices and edges are added.
Since the result of the optimization is a simple graph too, no further preprocessing
is needed to perform a subsequent optmization.

ergl:
Actually, I'm not too happy about this implementation. I'll push some changes.

ergl:
Ok, now binary functions are merged together:



Merging unary with binary doesn't seem so easy, so the result is the same
as last time.



However, if the child vertex has more than two parents, no optimization
will be attempted, and multiple edges will be allowed. If another binary
function is added, it will just be created as an extra edge. Unary functions
will always be merged together, though.

ergl:
Ok, I actually fear this might be getting too complex. I'd rather keep just the unary optimization
because the binary merging is proving to bee too much of a hassle

ergl:
Alright, I think this PR is done

src/lasp_dependence_dag.erl
```diff
@@ -318,6 +320,89 @@ get_direct_edges(G, V1, V2) ->
...
+    %% If we have less than 2 edges, we are already optimized.
+
+    Read = {Src, L#edge_label.read},
+
+    %% Gather all transform functions with their associated Pids.
+    {Transforms, Pids} = collect_transform_functions(Edges),
+
+    Write = {Dst, L#edge_label.write},
+
+    NewTransform = fun(V) ->
```
cmeiklejohn: I'm at a loss for what this is actually doing.  Can you clarify?

src/lasp_dependence_dag.erl
```diff
@@ -318,6 +320,89 @@ get_direct_edges(G, V1, V2) ->
...
+        orset_union_list(lists:map(fun(F) -> F(V) end, Transforms))
```
cmeiklejohn: Why OR-Set union?  I don't understand what's happening here?

cmeiklejohn: Is the general idea here that you're going to return a list of results and then union them together?

src/lasp_dependence_dag.erl
```diff
@@ -318,6 +320,89 @@ get_direct_edges(G, V1, V2) ->
...
+    end, H, T).
+
+%% @doc Kill all given Pids and then start a new lasp process.
+%%
+%%      Old Pids that are killed will remove themselves from
+%%      the dag, and the new one will register automatically.
+%%
+%%      Execute in a new process, otherwise we block.
+%%
+kill_then_start(Pids, Args) ->
```
cmeiklejohn: I'm concerned about race conditions here with a new process spawning in between the processes terminating and restarting.

src/lasp_dependence_dag.erl
```diff
@@ -318,6 +320,89 @@ get_direct_edges(G, V1, V2) ->
...
+%%      new lasp process with the mentioned function.
+%%
+%%      Note: assumes all read and write functions are the same in all
+%%      edges.
+%%
+%%      @todo add support for deltas (see map:core/6)
+%%
+-spec union_unary([edge()]) -> ok | pid().
+union_unary([{_, Src, Dst, L}, _ | _]=Edges) ->
+    %% If we have less than 2 edges, we are already optimized.
```
cmeiklejohn: Can you clarify why this is true?

ergl: `union_unary` is called with the edges that represent unary functions between two
vertices. As the goal of this is to reduce this number of edges to one, it only kicks
in when we have 2 or more edges.

cmeiklejohn: Can you add a comment to the code for future readers?

src/lasp_dependence_dag.erl
cmeiklejohn: This needs to be at least parameterized on the type of object it's getting -- for instance, avoid the optimization if this is a G-Counter, right?

src/lasp_dependence_dag.erl
ergl: Done

src/lasp_dependence_dag.erl
ergl: You're right, although looking through the implementations of map, filter and the set-theoretic functions, they all assume that they are getting a `state_orset`, maybe that should change as well?

src/lasp_dependence_dag.erl
ergl: I'm looking into this

ergl: You're right, something like this

```erlang
spawn(fun() ->
    spawn(fun() ->
        lasp:map(S1, fun(X) -> X * 2 end, S2)
    end),
    spawn(fun() ->
        lasp_dependence_dag:merge_unary_edges(S1, S2)
    end)
end).
```

results in two processes between S1 and S2 (doesn't matter the order of the internal processes).

Is there something we can do to fix this?

cmeiklejohn:
I'm still not convinced that this optimization is sound.

* You need to handle the case of other data types -- for example, the `bind_to` function for all types (G-Counter, PN-Counter, etc.) is going to use a process to drive the mapping between two variables, which does not seem accounted for in this code.
* I'm not convinced `union` is right.  If you have multiple edges into the same vertex from any source with different types of transformations the result is not *unioned* (in, the causal metadata sense), it's *merged* by the backend.  Now, thinking this through further probably highlights the fact that the result of the merge operation is non-sensical in many cases because of this behavior, so maybe we need to prohibit the ability to create these edges in the first place.

ergl:
You're right. I'm going to close this for now until I find a better way to do it

PR #119, opened 2016-06-26
Instrumentation

vitorenesduarte:
Integration branch for evaluation wired version of Lasp.


src/lasp_advertisement_counter.erl
```diff
@@ -360,10 +360,10 @@ servers(SetType, Ads, AdsWithContracts) ->
                 end, AdList).
 
 %% @private
-log_transmission(Instrumentation, Term) ->
+log_transmission(Instrumentation, Payload) ->
     case Instrumentation of
         true ->
-            lasp_transmission_instrumentation:log(client, Term, node());
+            lasp_transmission_instrumentation:log(client, from_server, Payload, 1, node());
```
cmeiklejohn: What designates the server here?  You're not going to be able to know this at this level, I think.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -1005,11 +1005,11 @@ do(Function, Args) ->
 -endif.
 
 %% @private
-log_transmission(Term) ->
+log_transmission(Type, Payload, PeerCount) ->
     try
         case lasp_config:get(instrumentation, false) of
             true ->
-                ok = lasp_transmission_instrumentation:log(server, Term, node()),
+                ok = lasp_transmission_instrumentation:log(server, Type, Payload, PeerCount, node()),
```
cmeiklejohn: Again, I'm not sure what ```server``` means here, as this function can be triggered by any node in the system -- one of which will be acting as a client and server both.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -1038,7 +1038,8 @@ schedule_delta_garbage_collection() ->
     end.
 
 %% @private
-send(Msg, Peer) ->
+send({Type, _From, Payload}=Msg, Peer) ->
```
cmeiklejohn: In addition to send, you're also going to need this instrumentation in both ```delta_send``` and ```delta_ack```.

src/lasp_transmission_instrumentation.erl
```diff
@@ -40,9 +40,9 @@
 -include("lasp.hrl").
 
 %% State record.
--record(state, {type,
+-record(state, {entity,
                 tref,
-                size=0,
+                size_per_type=orddict:new(),
```
cmeiklejohn: What is this used for?

src/lasp_transmission_instrumentation.erl
```diff
@@ -57,53 +57,59 @@
...
-    gen_server:call({global, {?MODULE, Type}}, {log, Term, Node}, infinity).
+-spec log(term(), term(), term(), pos_integer(), node()) -> ok | error().
+log(Entity, Type, Payload, PeerCount, Node) ->
+    gen_server:call({global, {?MODULE, Entity}}, {log, Type, Payload, PeerCount, Node}, infinity).
 
 -spec start(term(), list(), pos_integer()) -> ok | error().
-start(Type, Filename, Clients) ->
-    gen_server:call({global, {?MODULE, Type}}, {start, Filename, Clients}, infinity).
+start(Entity, Filename, Clients) ->
+    gen_server:call({global, {?MODULE, Entity}}, {start, Filename, Clients}, infinity).
```
cmeiklejohn: This can't be global.

src/lasp_transmission_instrumentation.erl
```diff
@@ -57,53 +57,59 @@
...
-    gen_server:start_link({global, {?MODULE, Type}}, ?MODULE, [Type], []).
+start_link(Entity) ->
+    gen_server:start_link({global, {?MODULE, Entity}}, ?MODULE, [Entity], []).
 
--spec log(term(), term(), node()) -> ok | error().
-log(Type, Term, Node) ->
-    gen_server:call({global, {?MODULE, Type}}, {log, Term, Node}, infinity).
+-spec log(term(), term(), term(), pos_integer(), node()) -> ok | error().
+log(Entity, Type, Payload, PeerCount, Node) ->
+    gen_server:call({global, {?MODULE, Entity}}, {log, Type, Payload, PeerCount, Node}, infinity).
```
cmeiklejohn: This can't be global.

src/lasp_transmission_instrumentation.erl
```diff
@@ -57,53 +57,59 @@
 
 %% @doc Start and link to calling process.
 -spec start_link(list(term()))-> {ok, pid()} | ignore | {error, term()}.
-start_link(Type) ->
-    gen_server:start_link({global, {?MODULE, Type}}, ?MODULE, [Type], []).
+start_link(Entity) ->
+    gen_server:start_link({global, {?MODULE, Entity}}, ?MODULE, [Entity], []).
```
cmeiklejohn: This can't be global.

src/lasp_transmission_instrumentation.erl
```diff
@@ -57,53 +57,59 @@
...
-start(Type, Filename, Clients) ->
-    gen_server:call({global, {?MODULE, Type}}, {start, Filename, Clients}, infinity).
+start(Entity, Filename, Clients) ->
+    gen_server:call({global, {?MODULE, Entity}}, {start, Filename, Clients}, infinity).
 
 -spec stop(term()) -> ok | error().
-stop(Type) ->
-    gen_server:call({global, {?MODULE, Type}}, stop, infinity).
+stop(Entity) ->
+    gen_server:call({global, {?MODULE, Entity}}, stop, infinity).
```
cmeiklejohn: This can't be global.

src/lasp_plumtree_broadcast_distribution_backend.erl
vitorenesduarte: `lasp_transmission_instrumentation` stored something called `type` (that I renamed to `entity`). The possible values were: `client` and `server`.
Are you saying we don't need this distinction anymore?

src/lasp_plumtree_broadcast_distribution_backend.erl
vitorenesduarte: This function is used for `aae_send`, `delta_send` and `delta_ack`

src/lasp_transmission_instrumentation.erl
vitorenesduarte: The idea is to count the number of bytes sent per type:

- `aae_send`
- `delta_ack`
- `delta_send`
- `broadcast`
- ...

src/lasp_transmission_instrumentation.erl
vitorenesduarte: Just changing to `local` didn't work.
And the error was not clear to me, so I might need help on this one.

src/lasp_plumtree_broadcast_distribution_backend.erl
cmeiklejohn: That distinction was so it could run twice on the same server and create two different log files because they each used a different way of tracking transmission.  This is no longer needed, given each instance running during the evaluation has a different purpose.  I'd just remove it, as I think it's no longer needed.

src/lasp_plumtree_broadcast_distribution_backend.erl
cmeiklejohn: Ah, perfect.  Nevermind, then!

src/lasp_transmission_instrumentation.erl
cmeiklejohn: That sounds good to me.

How will you distinguish in the output log file that you generate at the end of the execution?

src/lasp_transmission_instrumentation.erl
cmeiklejohn: First, start by removing Entity and Type, because those attributes are not required.

Then, you should be able to change it to `{local, ?MODULE}`.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -177,8 +177,7 @@ broadcast_data(#broadcast{id=Id, type=Type, clock=Clock,
 %%      for the same object.
 -spec merge({broadcast_id(), broadcast_clock()}, broadcast_payload()) ->
     boolean().
-merge({Id, Clock}, {Id, Type, Metadata, Value}=Payload) ->
-    log_transmission(Payload),
```
vitorenesduarte: @cmeiklejohn this removal makes sense right?

cmeiklejohn: Yes, that's correct.

src/lasp_plumtree_broadcast_distribution_backend.erl
```diff
@@ -897,11 +896,12 @@ code_change(_OldVsn, State, _Extra) ->
 
 %%%===================================================================
 %%% Internal functions
-%%%===================================================================
+%%%===============================1====================================
```
cmeiklejohn: This looks like a typo.

vitorenesduarte: My vim skills just keep surprising me.

src/lasp_transmission_instrumentation.erl
```diff
@@ -60,12 +59,9 @@ start_link() ->
 
 -spec log(term(), term(), pos_integer(), node()) -> ok | error().
 log(Type, Payload, PeerCount, Node) ->
+    lager:info("lasp_transmission_instrumentation:log"),
```
cmeiklejohn: Might want to pull these out, as they do slow down the execution of the code.

cmeiklejohn:
- [ ] Mechanism for disabling the broadcast tree for all modes.
- [ ] State-based AAE (no broadcast tree) graph with a line (total AAE) per node.
- [ ] State-based AAE (with broadcast tree) graph with a line (total AAE + broadcast tree) per node.
- [ ] Delta-based AAE (no broadcast tree) graph with a line (total AAE) per node.
- [ ] Graph with one line total for all nodes with three lines: state AAE, state AAE + tree, delta AAE.

src/lasp_advertisement_counter_client.erl
```diff
@@ -100,15 +100,16 @@ handle_info(log, #state{impressions=Impressions}=State) ->
     schedule_logging(),
 
     {noreply, State};
-handle_info(view, #state{actor=Actor, impressions=Impressions}=State0) ->
+
+handle_info(view, #state{actor=Actor, impressions=Impressions0}=State) ->
```
cmeiklejohn: Don't you need to initialize this at zero when the system starts?

vitorenesduarte: https://github.com/lasp-lang/lasp/blob/instrumentation/src/lasp_advertisement_counter_client.erl#L60-L73

I think this does that. No?

vitorenesduarte: `impression` was already there.
I'm just stopping the schedule if it reaches max.

PR #124, opened 2016-07-03
Integrate the oorset type

junghunyoo:


src/lasp_core.erl
```diff
@@ -361,7 +361,7 @@ update(Id, Operation, Actor, Store) ->
     {ok, var()} | not_found().
 update(Id, Operation, Actor, MetadataFun, Store) ->
     {ok, #dv{value=Value0, type=Type}} = do(get, [Store, Id]),
-    {ok, Value} = lasp_type:update(Type, Operation, Actor, Value0),
+    {ok, Value} = lasp_type:update(Type, Operation, {Id, Actor}, Value0),
```
cmeiklejohn: Why is this change required?

src/lasp_type.erl
```diff
@@ -101,15 +102,24 @@ new(Type) ->
...
-            T:delta_mutate(Operation, Actor, Value);
+            T:delta_mutate(Operation, RealActor, Value);
         state_based ->
-            T:mutate(Operation, Actor, Value);
+            T:mutate(Operation, RealActor, Value);
         pure_op_based ->
             ok %% @todo
     end.
 
+%%private
```
cmeiklejohn: Should be ```@private```

junghunyoo: I will change it.

src/lasp_core.erl
junghunyoo: I copied my comment in types PR.

In order to handle the union()/intersection()/product() operations properly, the actor identifier was not enough. Therefore, I created the EventId which is {ObjectId, ActorId}. If this is not allowed, then I have to find another way for it since this is an essential requirement for those operations.

cmeiklejohn: Does this need to be unique?

cmeiklejohn: Rather, a better question is *how* unique does this need to be?  Is this what's being used for `next()` in the semantics that you mailed over earlier?

junghunyoo: It should be globally unique. That is why I paired the ObjectId and the ActorId. I think the ActorId can only grantee the object level uniqueness, is this correct?

cmeiklejohn: Globally unique for an actor modifying an object or globally unique representing a globally unique event?

junghunyoo: I wanted a globally unique event in order to distinguish them for the binary set operations.

src/lasp_type.erl
```diff
@@ -101,15 +102,24 @@ new(Type) ->
...
+get_actor(state_oorset_ps, {{StorageId, _TypeId}, Actor}) ->
```
cmeiklejohn: Question -- could we just normalize across the board on using `{ObjectId, Actor}` for the actor identifiers to avoid any special casing for optimized CRDTs?

cc: @vitorenesduarte @CBaquero

vitorenesduarte: @cmeiklejohn Can you elaborate?

cmeiklejohn: Given that the optimized representations require an actor that's unique for a given object, like `{ObjectId, ActorId}`, I was thinking we could just standardize this in the backend for everything.

vitorenesduarte: So, for example, in a `GCounter`
instead of
`Actor -> Count`
we would have
`{ObjectId, ActorId} -> Count`
?

cmeiklejohn: Yes, that's correct.  This would be a change directly in Lasp only, just for uniform treatment of objects in the system.

src/state_orset_ext.erl
```diff
@@ -27,15 +27,130 @@
          product/2,
          filter/2]).
 
-union(LValue, RValue) ->
-    state_orset:merge(LValue, RValue).
+union({Type, LValue}, {Type, RValue}) ->
+    lager:info("union"),
```
cmeiklejohn: Can you remove this extra logging?

src/state_orset_ext.erl
```diff
@@ -51,12 +166,75 @@ intersect_folder({state_orset, RValue}) ->
             {state_orset, Acc ++ Values}
     end.
 
+map(Function, {state_oorset_ps, {{ElemDataStore, EventDataStore},
+                                 FilteredOutEvents, AllEvents}}) ->
+    lager:info("map"),
```
cmeiklejohn: Same.

PR #131, opened 2016-07-08
Track ephemerals (wip, don't merge)

ergl:
Track `read` and `bind` in the dag. `query` is tracked too since it doesn't use `read`.

~~Old lasp processes use non-tracked versions (`read_var` and `bind_var`), since I
couldn't get them to work with the tracked versions.~~

~~Todo: track `read_any`~~

### DIFF
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -205,7 +205,7 @@ read(Id, Threshold, Store) ->
             {error, Reason};
 
         block -> receive
-            X -> X
+            {ok, {_Id, _Type, _Metadata, _Value}}=Res -> Res
         end
     end.
 
--- a/src/lasp_core.erl
+++ b/src/lasp_core.erl
@@ -69,11 +69,6 @@
          intersection/7,
          fold/6]).
 
-%% Non-tracked versions of read / write
-%% @todo remove these, use only tracked versions
--export([bind_var/3,
-         read_var/3]).
-
 %% Administrative controls.
 -export([storage_backend_reset/1]).
 
@@ -84,12 +79,12 @@
 
 -define(WRITE, fun(_Store) ->
                  fun(_AccId, _AccValue) ->
-                   {ok, _} = ?MODULE:bind_var(_AccId, _AccValue, _Store)
+                   {ok, _} = ?MODULE:bind(_AccId, _AccValue, _Store)
                  end
                end).
 
 -define(READ, fun(_Id, _Threshold) ->
-                ?MODULE:read_var(_Id, _Threshold, Store)
+                ?MODULE:read(_Id, _Threshold, Store)
               end).
 
 %% @doc Initialize the storage backend.
@@ -190,17 +185,16 @@ read(Id, Store) ->
 %%
 -spec read(id(), value(), store()) -> {ok, var()}.
 read(Id, Threshold, Store) ->
-    Self = self(),
-    ReplyFun = fun({Id1, Type, Metadata, Value}) ->
-                       {ok, {Id1, Type, Metadata, Value}}
-               end,
-    BlockingFun = fun() ->
-                receive
-                    X ->
-                        X
-                end
-            end,
-    read(Id, Threshold, Store, Self, ReplyFun, BlockingFun).
+    ReplyFun = fun
+        ({error, _}) -> error;
+        ({_Id, _Type, _Metadata, _Value}=Found) ->
+            {ok, Found}
+    end,
+    BlockingFun = fun() -> block end,
+    case read(Id, Threshold, Store, self(), ReplyFun, BlockingFun) of
+        {ok, FoundValue} -> {ok, FoundValue};
+        _ -> receive X -> X end
+    end.
 
 %% @doc Perform a monotonic read for a series of given idenfitiers --
 %%      first response wins.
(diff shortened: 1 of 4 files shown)


src/lasp_core.erl
```diff
@@ -556,45 +556,23 @@ read_var(Id, Threshold0, Store, Self, ReplyFun, BlockingFun) ->
 %% @doc Perform a read (or monotonic read) for a series of particular
 %%      identifiers.
 %%
-%%      @todo track in dag
 %%
 -spec read_any([{id(), value()}], pid(), store()) ->
     {ok, var()} | {ok, not_available_yet}.
 read_any(Reads, Self, Store) ->
```
ergl: I assume this change is ok, I looked into the implementation of `read_any` in the [plumtree module](https://github.com/lasp-lang/lasp/blob/unstable/src/lasp_plumtree_broadcast_distribution_backend.erl#L364) and it was already using the `read` function

src/lasp_core.erl
```diff
@@ -182,20 +182,23 @@ read(Id, Store) ->
...
-               end,
-    BlockingFun = fun() ->
-                receive
-                    X ->
-                        X
-                end
-            end,
-    read(Id, Threshold, Store, Self, ReplyFun, BlockingFun).
+    ReplyFun = fun
+        ({error, _}) -> error;
```
cmeiklejohn: Why not return `{error, Reason}`?  You've got a reason here, and it's probably good to propagate it forward, at least for logging.

src/lasp_core.erl
```diff
@@ -182,20 +182,23 @@ read(Id, Store) ->
...
+        ({_Id, _Type, _Metadata, _Value}=Found) ->
+            {ok, Found}
+    end,
+    BlockingFun = fun() -> block end,
```
cmeiklejohn: I don't understand the significance of this change.

src/lasp_core.erl
```diff
@@ -182,20 +182,23 @@ read(Id, Store) ->
...
+    case read(Id, Threshold, Store, self(), ReplyFun, BlockingFun) of
+        {ok, FoundValue} -> {ok, FoundValue};
+        _ -> receive X -> X end
```
cmeiklejohn: Is the only possible value for this ignore `block`?

src/lasp_core.erl
ergl: You're right, I'll change that

src/lasp_core.erl
ergl: Yes, it should only `receive` on `block`, I'll make it explicit and add a match for the error

zeeshanlakhani: makes sense.

src/lasp_dependence_dag.erl
```diff
@@ -214,14 +214,10 @@ handle_call({will_form_cycle, From, To}, _From, #state{dag=Dag}=State) ->
...
 
-    %% All other vertices are tracked at the `declare` level, but
-    %% stream is a special case. We check if it is already a vertex
-    %% in the graph since adding an already existing vertex removes
-    %% any previous vertex labels.
-    case (Dst =:= stream) andalso (digraph:vertex(Dag, Dst) =:= false) of
-        true -> digraph:add_vertex(Dag, Dst);
-        _ -> ok
-    end,
+    %% Add vertices only if they are either top or bottom. (See add_if)
```
cmeiklejohn: What does `top` or `bottom` mean here?

src/lasp_dependence_dag.erl
```diff
@@ -326,6 +322,35 @@ get_direct_edges(G, V1, V2) ->
...
+    case Test(V) andalso (digraph:vertex(Dag, V) =:= false) of
+        true -> digraph:add_vertex(Dag, V);
+        _ -> ok
+    end.
+
+is_bottom_vertex(V) ->
+    lists:member(V, [stream, read, query]).
+
+is_top_vertex(bind) -> true;
+is_top_vertex(_) -> false.
```
cmeiklejohn: How is `update` treated?  Is that just a specific case of `read` and `bind`?

src/lasp_process.erl
```diff
@@ -76,6 +77,40 @@ start_tracked_process(EventCount, [ReadFuns, TransFun, {To, _}=WriteFun]) ->
...
+%%      create a single-fire lasp process representing the dataflow
+%%      computation.
+%%
+%%      This function is synchronous, as it waits for a value to
+%%      be returned.
+%%
+single_fire_function(From, To, Fn, Args) ->
+    Self = self(),
+    %% Unique reference, ignore any other messages.
+    %% @todo Remove?
```
cmeiklejohn: I wouldn't remove, the reference seems useful here for matching only the message that you are expecting.

src/lasp_dependence_dag.erl
ergl: Yes, `update` works the same, it triggers a `bind` underneath. The dag will get notified that a given variable is being updated with a `bind`

cmeiklejohn:
:+1:, looks good to merge.

src/lasp_dependence_dag.erl
ergl: It just means vertices that can't have parents (top) or children (bottom).
A better name for top is source vertex and sink for bottom, although I don't know

cmeiklejohn: Yeah, I'm thinking source and sink are better terms for these two.

ergl: Alright, let me change that

PR #132, opened 2016-07-08
Multi-mode graph

vitorenesduarte:
WIP
- [x] Move lasp_plot_gen.erl to an escript
- [x] Per mode, average all executions
- [x] Single graph with all modes (one graph with PS, one without)


cmeiklejohn:
This should eventually move to the evaluation repository, right?

vitorenesduarte:
Yes. Just while I'm working on it.

PR #138, opened 2016-07-23
Dag path contraction

ergl:
This PR implements path contraction and cleaving as a way to remove intermediate vertices in the graph.

#### What's working:

- Detection of suitable paths to be contracted.

- Explicit optimization passes (calling `lasp_dependence_dag:contract()` will contract all suitable paths in the graph).

- Automatic cleaving based on vertex access behaviours. (Creating new edges involving removed vertices undoes any optimizations involving that vertex).

- Cycle detection on removed edges (`lasp_dependence_dag:will_form_cycle` will detect cycles even if the vertices were removed during a contraction in the graph).

- Autoupdating of vertex tags on process restart.

- Automatic contraction passes.

#### To be implemented:

- [X] ~~Removed vertices hold a reference to the Pid of the process that connects endpoints in the contracted path. This reference is not updated if that process is restarted. This means that if that process is restarted, any future attempts to cleave the vertices in that path will fail.~~

- [X] ~~Automatic optimization passes. Think about how often optimization passes should occur.~~

#### Issues

- Since cleavings happen _after_ an edge is created, reads and queries may be able to get old data from a removed vertex. This is not a problem with binds or updates since downstream values will eventually get the new value once the cleaving finishes.

### DIFF
--- a/src/lasp_dependence_dag.erl
+++ b/src/lasp_dependence_dag.erl
@@ -752,11 +752,12 @@ contract(G, VSeq, State) ->
     {ok, NewState} = add_edges(nostep, [First], Last, Pid,
                                [Read], TransFun, Write, State),
 
-    OptMap = NewState#state.optimized_map,
+    NewDag   = NewState#state.dag,
+    OptMap   = NewState#state.optimized_map,
     PidTable = NewState#state.pid_table,
 
     %% Remove the intermediate edges by terminating the associated processes.
-    NewOptMap = remove_edges(G, VSeq, ProcessHash, Pid, OptMap),
+    NewOptMap = remove_edges(NewDag, VSeq, ProcessHash, Pid, OptMap),
 
     NewState#state{optimized_map = NewOptMap, pid_table = dict:store(Pid, ProcessHash, PidTable)}.
 
--- a/src/lasp_dependence_dag.erl
+++ b/src/lasp_dependence_dag.erl
@@ -39,8 +39,7 @@
          code_change/3]).
 
 %% Defines how often an optimization pass happens.
-%% A value of 0 means the optimization happens every time.
--define(CONTRACTION_INTERVAL, 0).
+-define(CONTRACTION_INTERVAL, 1000).
 
 %%%===================================================================
 %%% Type definitions
@@ -92,12 +91,6 @@
                                     contract_path(),
                                     list(#process_metadata{})}).
 
-%% A step is defined as a call to add_edges. While we are defining our
-%% application topology, the graph is suitable to be changed regularly.
-%% See ?CONTRACTION_INTERVAL. One the step counter reaches that value,
-%% an optimization pass occurs, and then is reset to zero.
--type contraction_step() :: non_neg_integer().
-
 %% Only used to represent paths suitable to be contracted. These paths
 %% consist of two necessary endpoints and a list of unnecessary vertices
 %% in between.
@@ -117,7 +110,7 @@
                 process_map :: process_map(),
                 optimized_map :: optimized_map(),
                 pid_table :: pid_table(),
-                contraction_step :: contraction_step()}).
+                contraction_timer :: timer:tref()}).
 
 %% We store the function metadata as the edge label.
 %% An edge in the graph represents a lasp process, and contains
@@ -263,11 +256,19 @@ cleave_all() ->
 
 %% @doc Initialize state.
 init([]) ->
+    Timer = case lasp_config:get(automatic_contraction, false) of
+        true ->
+            {ok, Tref} = timer:send_after(?CONTRACTION_INTERVAL, contract),
+            Tref;
+
+        _ ->
+            undefined
+    end,
     {ok, #state{dag=digraph:new([acyclic]),
                 process_map=dict:new(),
                 optimized_map=dict:new(),
                 pid_table=dict:new(),
-                contraction_step=0}}.
+                contraction_timer=Timer}}.
 
 -spec handle_call(term(), {pid(), term()}, #state{}) ->
     {reply, term(), #state{}}.
@@ -309,10 +310,8 @@ handle_call({export_dot, Path}, _From, #state{dag=Dag}=State) ->
 handle_call(get_process_map, _From, #state{process_map=PM}=State) ->
     {reply, {ok, dict:to_list(PM)}, State};
 
-handle_call(contract, _From, #state{dag=Dag}=State) ->
-    NewState = lists:foldl(fun(P, S) ->
-        contract(Dag, P, S)
-    end, State, contraction_paths(Dag)),
+handle_call(contract, _From, State) ->
+    NewState = contract_all(State),
     {reply, ok, NewState};
 
 handle_call({cleave, Vertex}, _From, #state{dag=Dag, optimized_map=OptMap}=State) ->
@@ -365,7 +364,7 @@ handle_call({will_form_cycle, From, To}, _From, #state{dag=Dag, optimized_map=Op
 %%      We monitor all edge Pids to know when they die or get restarted.
 %%
 handle_call({add_edges, Src, Dst, Pid, ReadFuns, TransFun, WriteFun}, _From, State) ->
-    {Reply, NewState} = add_edges(step, Src, Dst, Pid,
+    {Reply, NewState} = add_edges(Src, Dst, Pid,
                                   ReadFuns, TransFun, WriteFun, State),
     {reply, Reply, NewState}.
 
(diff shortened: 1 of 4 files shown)


src/lasp_dependence_dag.erl
```diff
@@ -24,6 +24,11 @@
          out_edges/1,
          in_edges/1]).
 
+%% @todo Only export on test. Optimization Debug
```
cmeiklejohn: Wrap this in an ```-ifdef(TEST).```

src/lasp_dependence_dag.erl
```diff
@@ -44,20 +53,53 @@
 %% lets us quickly delete those edges.
 -type process_map() :: dict:dict(pid(), {id(), id()}).
 
+%% Stored metadata for a lasp process.
+%% We don't store the input(s) and output vertices as that information
+%% is implicitly stored by edges in the graph.
+-record(process_metadata, {read :: function(),
```
cmeiklejohn: Does this mean that a single process only read and writes to *one* object, given that it's a single function?

src/lasp_dependence_dag.erl
```diff
@@ -44,20 +53,53 @@
...
+                           transform :: function(),
+                           write :: function()}).
+
+%% We store a mapping Id -> {Pid, Path, MetadataList} to identify the list of
+%% vertices and relationships that get removed as part of a path contraction.
+%%
+%% Used during the cleaving step.
+-type contract_process_id() :: non_neg_integer().
```
cmeiklejohn: Can you clarify specifically what this is?

cmeiklejohn: Also, why isn't this a `pid()`?

src/lasp_dependence_dag.erl
```diff
@@ -123,14 +171,26 @@ in_edges(V) ->
 process_map() ->
     gen_server:call(?MODULE, get_process_map, infinity).
 
+contract() ->
```
cmeiklejohn: Can you add a doc string for this?

src/lasp_dependence_dag.erl
```diff
@@ -123,14 +171,26 @@ in_edges(V) ->
...
+    gen_server:call(?MODULE, contract, infinity).
+
+cleave(Vertex) ->
```
cmeiklejohn: Can you add a doc string for this?

src/lasp_dependence_dag.erl
```diff
@@ -123,14 +171,26 @@ in_edges(V) ->
...
+    gen_server:call(?MODULE, {cleave, Vertex}, infinity).
+
+cleave_all() ->
```
cmeiklejohn: Can you add a doc string for this?

src/lasp_dependence_dag.erl
```diff
@@ -176,6 +232,25 @@ handle_call({export_dot, Path}, _From, #state{dag=Dag}=State) ->
 handle_call(get_process_map, _From, #state{process_map=PM}=State) ->
     {reply, {ok, dict:to_list(PM)}, State};
 
+handle_call(contract, _From, #state{dag=Dag}=State) ->
+    lists:foreach(fun(P) ->
```
cmeiklejohn: Is there any possibility that one of these operations can fail?

src/lasp_dependence_dag.erl
ergl: Yes, this metadata is only saved for process that are removed during a path contraction,
meaning they link unnecessary vertices only. That's why they only have one read function.

Processes that have _n_ inputs are represented by _n_ edges, each one from one of the inputs to the output, with the appropiate read function as part of the edge label.

src/lasp_dependence_dag.erl
ergl: I don't think so, I haven't found any errors so far

src/lasp_dependence_dag.erl
ergl: At first I was using the pid directly, but if the process is restarted by a supervisor, we have no way of knowing the new pid in order to update this information.

However, as far as I know, when a process is restarted, the supervisor uses exactly the same `init` arguments every time. This `contract_process_id` is just a hash of the read, transform and write functions so we can keep track of processes through restarts.

When a new edge is added to the graph, we check if we have previously seen the hash. If we have, then that process is a restart of an older one, and we know that we have to update the optimized map.

The `pid_table` is there to map pids to its hash. When a process is killed, we only know
its pid, so we have to get their hash in order to correctly remove the appropiate metadata from the `optimized_map`.

test/lasp_SUITE.erl
```diff
@@ -103,6 +104,55 @@ all() ->
...
+            Threshold = case Threshold0 of
+                undefined -> {strict, undefined};
+                _ -> Threshold0
+            end,
+            MutateAndRead = fun(F, T, M, Th) ->
+                lasp:update(F, M, a),
+                lasp:read(T, Th)
+            end,
+            {Time, {ok, {_, _, _, NewThreshold}}} = timer:tc(MutateAndRead, [From, To, Mutator, Threshold]),
+            RC(Iterations - 1, [Time | Acc], From, To, {add, N+1}, NewThreshold)
```
vitorenesduarte: Here

cmeiklejohn: Roger

src/lasp_dependence_dag.erl
cmeiklejohn: I'd add a comment in the code to make that a bit clearer for the reader.

src/lasp_dependence_dag.erl
```diff
@@ -44,20 +54,53 @@
...
 
+%% @todo Move somewhere else
+%% A tuple of the arguments of a lasp process.
+-type process_args() :: {[{lasp_vertex(), function()}],
+                         function(),
+                         {lasp_vertex(), function()}}.
+
+-record(vertex_label, {process_pointer :: contract_process_id()}).
+
+-type lasp_vertex() :: id() | pid().
```
cmeiklejohn: What this and what are these types?  Can this be documented?

src/lasp_dependence_dag.erl
cmeiklejohn: Make define a type for the hash and comment it to say how it's computed.  It's not straightforward in the code, but is one you explain it.

src/lasp_dependence_dag.erl
```diff
@@ -66,19 +109,19 @@
...
+-spec add_vertices(list(lasp_vertex())) -> ok.
 add_vertices([]) ->
     ok;
 
 add_vertices(Vs) ->
     gen_server:call(?MODULE, {add_vertices, Vs}, infinity).
 
 %% @doc Check if linking the given vertices will form a loop.
--spec will_form_cycle(list(id()), id()) -> boolean().
+-spec will_form_cycle(list(lasp_vertex()), lasp_vertex()) -> boolean().
```
cmeiklejohn: Can you add a comment in the code about when this is useful?

src/lasp_dependence_dag.erl
```diff
@@ -176,6 +256,25 @@ handle_call({export_dot, Path}, _From, #state{dag=Dag}=State) ->
...
+
+handle_call({cleave, Vertex}, _From, #state{dag=Dag, optimized_map=OptMap}=State) ->
+    cleave_if_contracted(Dag, Vertex, OptMap),
+    {reply, ok, State};
+
+handle_call(cleave_all, _From, #state{optimized_map=OptMap}=State) ->
+    lists:foreach(fun(Id) ->
+        {Pid, _, _} = dict:fetch(Id, OptMap),
+        spawn_link(fun() ->
+            lasp_process_sup:terminate_child(lasp_process_sup, Pid)
```
cmeiklejohn: I don't understand what's going on here -- why is this terminating processes?

src/lasp_dependence_dag.erl
ergl: Yeah, `contract_process_id` is supposed to be type for the hash, but perhaps it's not named accordingly. I can change that and put what I said above in a comment.

src/lasp_dependence_dag.erl
ergl: `process_args` is a shorthand for the function metadata in a lasp process, it's used to calculate the hash of a new process, and to restart new ones during the cleaving phase.

`vertex_label` is the label of the removed vertices, and contains a reference to the process that contracted the old path of a vertex.

`lasp_vertex` is a common name for the kinds of vertices in the graph, we can have normal crdt values (`id()`), or process pids for the reads, queries and binds.

I'll add these comments to the file.

src/lasp_dependence_dag.erl
ergl: To undo the optimization for a given path, it kills the process that connects its endpoints.

When any of those processes are killed by the superviros, the cleaving process is automatically started.

You can see the handler for that in [`handle_info`](https://github.com/ergl/lasp/blob/contraction/src/lasp_dependence_dag.erl#L426) and the actual cleaving in [`cleave_associated_path`](https://github.com/ergl/lasp/blob/contraction/src/lasp_dependence_dag.erl#L832).

PR #140, opened 2016-07-28
Add minimal support for a SQL interface on Lasp.

cmeiklejohn:


junghunyoo:
This looks good. If we can parameterise the identifier of the final output, it would be useful to share it with other processes.

cmeiklejohn:
@junghunyoo I'm confused -- the create call returns the identifier to the user?

junghunyoo:
In the ad counter test, we use the `<<"ads_with_contracts">>` for the id of shared object. I think it would be great if we could do similar behaviour with this SQL interface.

src/lasp_sql_materialized_view.erl
```diff
@@ -24,12 +24,36 @@
...
+         get_value/2,
+         create_table_with_values/2]).
 
 -record(state, {output_id, set_id, projection_output_id, predicate_output_id}).
 
 -define(DEFAULT, <<"hi">>).
 
+%% @doc Insert in a SQL table the given rows.
+create_table_with_values(TableName, Args) when is_atom(TableName) ->
+    Name = list_to_binary(atom_to_list(TableName)),
```
cmeiklejohn: Maybe use the `generate_identifier` function I defined in this file for this?

PR #142, opened 2016-07-28
Divergence

vitorenesduarte:
- [x] Make the client number configurable
- [x] Remove music festival example
- [x] Fix ORSet product
- [x] Change the way convergence is being tracked right now since it doesn't allow us to observe divergence
- [x] Log divergence in the end of ad_counter simulation
- [x] Make two test cases `peer_to_peer_divergence` and `client_server_divergence` and run them with different number of clients
- [x] Allow clients to also disable the ads (and add a third test case for that)
- [x] Make a graph: X (client number) and Y (divergence)


cmeiklejohn:
I'm not sure why we are getting the transient errors on Travis, but I believe this ready to be merged now.

PR #151, opened 2016-08-30
Push logs to S3

vitorenesduarte:


cmeiklejohn:
Did you test this?

vitorenesduarte:
No.

vitorenesduarte:
It assumes a bucket named __logs__.
Also, just noticed there might be a problem in the put_object part.
It's a list. 
Let me make that a string.

cmeiklejohn:
Can you also make sure to call create_bucket like we do in the sprinter code when it starts?

cmeiklejohn:
Yes, I believe put_object needs a binary.

cmeiklejohn:
Yes, it does require a binary.

PR #153, opened 2016-09-06
Instrumentation for memory utilization

vitorenesduarte:


src/lasp_default_broadcast_distribution_backend.erl
```diff
@@ -1150,6 +1143,36 @@ schedule_delta_garbage_collection() ->
...
+plumtree_memory_report() ->
+    PlumtreeBroadcast = erlang:whereis(plumtree_broadcast),
+    lager:info("Plumtree message queue: ~p",
+               [process_info(PlumtreeBroadcast, message_queue_len)]),
+    lager:info("Our message queue: ~p",
+               [process_info(self(), message_queue_len)]).
+
+%% @private
+memory_utilization_report() ->
+    TotalBytes = orddict:fetch(total, erlang:memory()),
```
cmeiklejohn: Is this bytes or words?

vitorenesduarte: > The first element Type is an atom describing memory type. The second element Size is the memory size in bytes.

http://erlang.org/doc/man/erlang.html#memory-0

I was assuming the transmission is the same
because I just reused the function that gives megabytes

vitorenesduarte: Now I see I could have used `memory/1`

cmeiklejohn:
Reviewed last night; looks good.

PR #156, opened 2016-09-10
Buffer updates to plumtree.

cmeiklejohn:
Plumtree can be easily overloaded; so periodically buffer updates and
broadcast them in an effort to reduce this problem.

### DIFF
--- a/src/lasp_default_broadcast_distribution_backend.erl
+++ b/src/lasp_default_broadcast_distribution_backend.erl
@@ -76,6 +76,7 @@
 %% State record.
 -record(state, {store :: store(),
                 actor :: binary(),
+                to_broadcast :: list(),
                 sync_counter :: non_neg_integer(),
                 counter :: non_neg_integer(),
                 gc_counter :: non_neg_integer()}).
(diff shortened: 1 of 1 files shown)


src/lasp_default_broadcast_distribution_backend.erl
```diff
@@ -502,12 +506,51 @@ init([]) ->
...
+                                           metadata=Metadata,
+                                           value=Value},
+                    {Time, _Result} = timer:tc(plumtree_broadcast,
+                                              broadcast,
+                                              [Broadcast, ?MODULE]),
+                    lasp_logger:extended("Time to broadcast: ~p", [Time]),
+                    ok
+              end, ok, ToBroadcast0),
+
+    %% Buffer latest update for that state.
```
junghunyoo: I think this comment is not correct. This is the point we flush the broadcast buffer.

junghunyoo:
The change looks very good.

src/lasp_default_broadcast_distribution_backend.erl
cmeiklejohn: You're right.

PR #158, opened 2016-09-11
Multiple Apps in DCOS && DCOS Runner

vitorenesduarte:
- [x] Add DCOS Runner Dockerfile and a starting script
- [x] Allow multiple experiments to run at the same time
- [x] Fix wait_for_connectedness
- [x] Fix Actor Id in ad counter experiment
- [x] When checking for connectedness, use client number as well
- [x] Once the graph is connected, log that time (it allow us to track the beginning of the experiment)
- [x] Add jitter to sprinter s3 calls
- [x] Make mailbox queue length logging configurable
- [x] Max total number of impressions is 9600 (see f4cf4f8b6b1b44e4d5bde5191447859002f05310)
- [x] In client-server, the server is not reactive anymore


cmeiklejohn:
:+1:

PR #164, opened 2016-09-27
Add object filter for convergence structure.

cmeiklejohn:
Allow the convergence function to synchronize when there are partitions,
by using a user supplied filter function for filtering out the objects
to synchronize.


cmeiklejohn:
Addresses #162.

vitorenesduarte:
:+1:

PR #165, opened 2016-09-27
Add reactive client/server.

cmeiklejohn:
Add a reactive client/server topology where messages are sent back with
the latest update when arrived.


cmeiklejohn:
Addresses #163.

Issue #161, opened 2016-09-27
Improve check for connectedness algorithm


vitorenesduarte:

cmeiklejohn:
How precisely should this be improved?  Can you put more information in the issue?

vitorenesduarte:
Instead of checking that every node is connected to every other node, I think we can:
- do a graph traversal (breath first e.g.) and
- if we visit all nodes and
- all the links were bidirectional (this requirement comes from symmetric views from HyParView, right?) then
- the graph is connected

With 512 clients, using `get_path` instead of `get_shortest_path`, still takes 3 min.

PR #166, opened 2016-09-27
New is_connected algorithm

vitorenesduarte:


vitorenesduarte:
Addresses #161

src/sprinter.erl
```diff
@@ -405,6 +388,30 @@ populate_graph(Name, Membership, Graph) ->
...
 
+breath_first(Root, Graph, Visited0) ->
+    %% Check if every link is bidirectional
+    %% If not, stop traversal
+    In = ordsets:from_list(digraph:in_neighbours(Graph, Root)),
+    Out = ordsets:from_list(digraph:out_neighbours(Graph, Root)),
+
+    Visited1 = ordsets:union(Visited0, [Root]),
+
+    case In == Out of
```
cmeiklejohn: I'm not sure I understand what the check is looking for?

vitorenesduarte: That the links are bidirectional

cmeiklejohn:
Can you comment the code a bit just to make it a bit clearer for the next reader as to what specifically is happening here?  Then I think we're probably good to merge it.

cmeiklejohn:
:+1: to merge.

cmeiklejohn:
Did this have some effect?

vitorenesduarte:
It's way faster, that was the motivation.
The algorithm we had before would take 8 minutes to check connectedness for 512 nodes

PR #171, opened 2016-10-03
Correctly configure the peer service.

cmeiklejohn:
The partisan_peer_service_manager is only ever read by partisan in partisan_config, not lasp_config.


vitorenesduarte:
:+1:

PR #176, opened 2016-11-02
Return list from tutorial query

gar1t:
While the data structure is a set, the tutorial code is designed to
simplify the use and understanding of the interfaces, not to be
correct from an API standpoint. Those who would use the library would
not presumably rely on a tutorial wrapper.


gar1t:
Um, retracting this for obvious reasons - not quite that simple :)

PR #177, opened 2016-11-02
Return tuple from tutorial:query/2 instead of sets

gar1t:
While the data structure is a set, the tutorial code is designed to
simplify the use and understanding of the interfaces, not to be
correct from an API standpoint. Those who would use the library would
not presumably rely on a tutorial wrapper.

A tuple in this case can be used to differentiate sets from sequences


cmeiklejohn:
Thanks :trollface:

PR #179, opened 2016-11-17
Allow separation of state transmission and protocol transmission logs

vitorenesduarte:


src/lasp_sup.erl
```diff
@@ -123,7 +123,7 @@ init(_Args) ->
     end,
 
     %% Configure Plumtree logging.
-    TransLogMFA = {lasp_instrumentation, transmission, [broadcast]},
+    TransLogMFA = {lasp_instrumentation, transmission},
```
cmeiklejohn: I think this, since it's a MFA, should be lasp_instrumentation, transmission, [] to make it at least match the module, function, arguments signature.

vitorenesduarte: I didn't know what MFA means.
Now it makes sense.

cmeiklejohn:
This appears to be failing?  Otherwise, looks good though.

PR #184, opened 2016-11-30
move erlcloud to just be in the release and rm unused plugin

tsloughter:
The `pc` plugin does not appear to be used, so I've removed it from the config. I moved the `rebar3_auto` plugin to a new profile `auto` to be used as `rebar3 as auto auto` (I know that is kind of annoying, better to put personal dev tools like that in your global rebar3 config).

I also moved `erlcloud` to the relx config instead of .app.src. This allows me to override the lasp deps from my project to remove erlcloud without it complaining.

Do you do your experiments by running the release? If so this should be fine, if not then I'll have write something to remove it from the .app.src file of lasp on my end instead.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -28,8 +28,6 @@
 {yrl_first_files, ["src/lasp_sql_parser.yrl"]}.
 
 {cover_enabled, true}.
-{plugins, [{pc, {git, "https://github.com/blt/port_compiler.git", {branch, "master"}}},
-           rebar3_auto]}.
 
 {erl_opts, [debug_info,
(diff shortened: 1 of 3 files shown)


cmeiklejohn:
Can you target Lasp's `unstable` branch?  Master is a bit out of date, and unstable is where most of the work is happening.

`pc` was there, mainly because we needed it to build eleveldb for the storage backend, but right now the storage backend is only ets and gets.

cmeiklejohn:
All of the experiments use `make stage` so, yes, they all use a release when running.

PR #189, opened 2016-12-07
New strategy for pruning a node from the Ack Map

vitorenesduarte:
`AckMap :: Node -> {Ack, GCCounter}`

- When sending a delta group, increment the `GCCounter`
- When receiving an ack, reset `GCCounter` to 0
- When GCing, only the entries in the map that have `GCCounter < MaxGCCounter` remain in the map.

Others:
- Fix non-atomic write in `receive_delta(Store, {delta_ack, Id, From, Counter})`
- Only send a delta-group/full-state when the last ack from a peer is less than the local delta counter; if equals then this peer has seen at least as much as we've seen; Also add a special condition so that clients in reactive server mode don't respect this
- Simplify `collect_deltas`


include/lasp.hrl
```diff
@@ -26,8 +26,7 @@
              metadata :: metadata(),
              delta_counter :: non_neg_integer(),
              delta_map :: orddict:orddict(),
-             delta_ack_map :: orddict:orddict(),
-             delta_eager_map = [] :: list(crdt())}).
```
vitorenesduarte: Yes. I didn't find any use for it.
What was the purpose of this map?

include/lasp.hrl
```diff
@@ -49,9 +48,13 @@
...
 
-%% @doc Garbage collection will happen after the certain number
-%%      (MAX_GC_COUNTER) of times of exchanges.
--define(MAX_GC_COUNTER, 7).
+%% @doc A node will be evicted from the Ack Map when the entry of
+%%      that node in the Ack Map remains unchanged for
+%%      (MAX_GC_COUNTER) exchanges.
+%%      If the delta exchange timer is 5 seconds and MAX_GC_COUNTER
+%%      is 20, a node will be evicted from the Ack Map after not
+%%      sending acks for ~2 minutes
```
vitorenesduarte: Yes. It really depends when then GC happens. It could be less than 2min, but it could also be more. Right?

cmeiklejohn: I guess I would just update the comment to be more precise.

cmeiklejohn: (Yes, it will be that time at a *minimum*, but given messages can be delayed in the mailbox, things might take a lot longer under load.)

include/lasp.hrl
cmeiklejohn: I have no clue, I had also made a note to have it removed.

PR #193, opened 2016-12-12
Apply join-decompositions to the received delta-buffer to remove redundant information

vitorenesduarte:


src/lasp_core.erl
```diff
@@ -466,9 +466,10 @@ bind_var(Origin, Id, Value, MetadataFun, Store) ->
                                 state_based ->
                                     {ok, WDT, Counter0, DeltaMap0};
                                 delta_based ->
+                                    Delta = lasp_type:delta(Type, state_driven, Value, Value0),
```
cmeiklejohn: What is `state_driven`?

cmeiklejohn: What does this function do?

vitorenesduarte: - `Value0` is the local state
- `Value` is the received state

This function extracts a `Delta` from `Value` such that:
- `Value0 \join Value = Value0 \join Delta` and `Delta` is minimum.

If `Value0 = {a, b, c}` and `Value = {a, d}`, then `Delta = {d}`.

vitorenesduarte: This received value can be a state-based CRDT or a digest of that CRDT (for that there's the `digest_driven` not yet implemented).
Here I think we'll always use `state_driven`.

But in the future we could use this when we're exchanging with a new peer:
- I can send my state and this peer will compute a delta so that we reach the same state (2 messages to convergence)
- I can send a digest of my state, this peer will compute a delta so that I reach its state. To reach convergence this has to be done bidirectionally (4 messages) or by piggy-backing the second digest (3 messages)

PR #194, opened 2016-12-13
New GSet simulation

vitorenesduarte:


.travis.yml
```diff
@@ -11,9 +11,10 @@ script:
   - make xref
   - make ct
   - make peer-to-peer-ad-counter-simulation
-  - make client-server-ad-counter-simulation
-  - make peer-to-peer-game-tournament-simulation
-  - make client-server-game-tournament-simulation
+  - make peer-to-peer-simple-simulation
+#  - make client-server-ad-counter-simulation
+#  - make peer-to-peer-game-tournament-simulation
```
cmeiklejohn: Anything we can do here to keep these still running?

Makefile
```diff
@@ -43,6 +43,8 @@ ct:
 ad-counter-simulations: client-server-ad-counter-simulation peer-to-peer-ad-counter-simulation ad-counter-overcounting ad-counter-partition-overcounting
 game-tournament-simulations: client-server-game-tournament-simulation peer-to-peer-game-tournament-simulation
 
+simple-simulations: peer-to-peer-simple-simulation
```
cmeiklejohn: Is there no client/server mode for this test?

bin/dcos-deploy.sh
```diff
@@ -92,7 +92,7 @@ cat <<EOF > lasp-server.json
   "ports": [0, 0],
   "env": {
     "LASP_BRANCH": "$LASP_BRANCH",
-    "AD_COUNTER_SIM_SERVER": "true",
+    "SIMPLE_SIM_SERVER": "true",
```
cmeiklejohn: Looks like this will permanently change the behavior of the docs-deploy script.  Maybe we should parameterize this from the environment instead and assume a default of false at the top of the file?

bin/transmission.sh
```diff
@@ -21,7 +21,7 @@ do
 done
 
 EVAL_NUMBER=1
-SIMULATION=ad_counter
+SIMULATION=simple
```
cmeiklejohn: Maybe make this configurable through the environment?

src/lasp_simple_client.erl
```diff
@@ -0,0 +1,188 @@
...
+handle_info(event, #state{actor=Actor,
+                          events=Events0}=State) ->
+    lasp_marathon_simulations:log_message_queue_size("event"),
+
+    {ok, TotalEvents} = lasp:query(?SIMPLE_BAG),
+
+    LocalEvents = case TotalEvents > 0 of
+        true ->
+            %% The server, once it detects connectedness,
+            %% will add one element to the bag.
```
cmeiklejohn: I'm confused.  Isn't a bag a multi set?

cmeiklejohn:
Subsumed by https://github.com/lasp-lang/lasp/pull/199.

PR #195, opened 2016-12-13
Plumtree modifications to reduce exponential overhead.

cmeiklejohn:
- [x] Single sender rooted tree, computed by the server node.
- [x] Have each gossip interval use the eager membership from plumtree.
- [x] Ensure that we omit protocol messages used to maintain other trees.
- [x] Empty message periodically gossiped from the server to maintain tree.
- [x] Break out plumtree broadcast behavior, it's no longer necessary.
- [x] Determine if the broadcast buffer is still necessary.
- [x] Cache servers in Sprinter.


cmeiklejohn:
Is this good to be merged?

PR #198, opened 2016-12-18
Cleanup.

cmeiklejohn:
- [ ] Replace AAE references in the evaluation repository.
- [x] Core needs to spawn_link the backend.
- [x] Break out backends.
- [x] Resolve waiting delta threads.
- [x] Make backend contain the extract call as a callback object.
- [ ] Verify through experiments.


cmeiklejohn:
@vitorenesduarte What changes are needed to the evaluation repository so this can be merged?

PR #199, opened 2016-12-19
[WIP] Back propagation.

cmeiklejohn:
- [ ] Switch labels to records to ensure no collision with Borja's label usage for automatic contraction.
- [x] Heavy clients, advertisement client counter.
- [x] Client-side code.
- [x] Topological sort on state-based dissemination.
- [ ] Topological sort on delta-based dissemination.
- [ ] Back propagate notifications.
- [ ] Investigate vector clock usage for each object.
- [ ] Map vector clock through.
- [x] Prohibit intermediary node modification.
- [x] Ensure advertisement counter doesn't require intermediary node modification to finish.
- [ ] Prevent computation if back propagation notification is received from node further up the graph.
- [x] What about terminal nodes?


cmeiklejohn:
Leaving unfinished and merging in just the DAG modifications we made.

PR #201, opened 2017-01-11
copy in mochiglobal and init config vars

tsloughter:
This is not as clean as the partisan PR. I didn't feel comfortable messing with `configure_defaults`... I think this is a good starting place for a future refactoring. A number of the config values appear related solely to simulations, which I'd hope one day can be removed or moved to a better location?

The use of OS vars would be better done by the release. If that is something you are open to I'd be happy to make that patch as well. It basically removes all of the `os:getenv` because the start script `relx` provides will replace env vars in `sys.config` with their values at runtime, so `application:gen_env` will result in the value from the environment.


cmeiklejohn:
What do you propose as the fix for the `os:getenv` calls?

cmeiklejohn:
We'll gladly take the `os:getenv` fixes if you can provide a patch for those, and you think it cleans up the code and makes it more readable.

tsloughter:
I realized we don't have a good way of doing defaults for unset os vars when using the `sys.config` method and replacing on release start. Each var would have to be set every time, so that isn't going to work. But makes me think I should look into improving relx to support that.

PR #210, opened 2017-01-29
Partisan changes.

cmeiklejohn:
- [x] Bump partisan to version that uses OS env
- [ ] Figure out why the test execution takes too long now


cmeiklejohn:
Do we need this branch?

PR #211, opened 2017-01-29
Investigate why new Partisan is significantly slower.

cmeiklejohn:


cmeiklejohn:
cda019f approximately the same as unstable, no real change other than some discrepancies from the additional RPC calls.

cmeiklejohn:
Adding @tsloughter for review.

cmeiklejohn:
- [ ] Test suite execution slows down dramatically when partisan is removed from included_applications.
- [ ] Why does Plumtree need to be an included_application?

cmeiklejohn:
not included_applications

$ pkill -9 beam.smp; time ./rebar3 ct -v -v --readable=false --suite=lasp_SUITE --case=uncontracted_latency_test

real    0m19.056s
user    0m4.186s
sys     0m0.873s

included_applications

real    0m20.651s
user    0m4.820s
sys     0m1.084s

PR #214, opened 2017-02-01
[WIP] Kubernetes.

cmeiklejohn:
- [x] Any where DCOS is mentioned has to be adapted.
- [x] Figure out if the lasp_marathon_simulations code has anything marathon specific.
- [x] Ensure the test suite executes correctly.


bin/kube.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+
+cd /tmp
+
+echo "Deleting all deployments"
+kubectl delete deployments --all
```
pigmej: why not to `delete -f` ?

cmeiklejohn: I dunno, because we're just messing around?

cmeiklejohn: Oh, I see what you mean now.

PR #215, opened 2017-02-15
Add blocking sync call that will store in progress syncs.

cmeiklejohn:
- [x] Store in progress syncs in a buffer so we don't immediately reply.
- [x] When a blocking sync request is received, the remote note must ack each object with it's node name and _From value.
- [x] When an ack is received, prune the map of values that are being blocked on.
- [x] When all values being blocked on are gone, then use gen_server reply to notify the waiting node.
- [x] Make configurable option for this setting in the update/bind calls.


src/lasp_state_based_synchronization_backend.erl
```diff
@@ -98,6 +166,13 @@ handle_cast({state_send, From, {Id, Type, _Metadata, Value}},
                                ?CLOCK_INCR(Actor),
                                ?CLOCK_INIT(Actor)}),
 
+    case AckRequired of
+        true ->
+            ?SYNC_BACKEND:send(?MODULE, {state_ack, node(), Id}, From);
```
vitorenesduarte: Maybe use `Actor` instead of `node()`

PR #216, opened 2017-02-16
Workflow V2.

cmeiklejohn:


cmeiklejohn:
[ ] Sprinter identifier needs both evaluation identifier and timestamp.
[ ] Workflow needs both evaluation identifier and timestamp.
[ ] Add labels to the kubernetes deployment.
[ ] Restrict deployment based on labels.
[ ] Only shutdown the right experiments.
[ ] Start multiple redis.
[ ] Ensure that the redis sync code pulls the right redis instance.

PR #218, opened 2017-02-20
split out test deps and move simulations out of src/

tsloughter:
This is a start towards cutting down the core lasp src to only what is needed (meaning keeping the simulations and test code out of the main src).

I moved deps for tests to the test profile and moved simulations related modules (I may have missed some) to `simulations/`.

There is still work to do with cleaning up `lasp_sup` and such, and I'd like to figure out getting erlcloud and webmachine out of the dependencies. 

I'm also not sure about the eredis dep, this may end up needing to stay in the main deps and included in the release like erlcloud is currently? If they are just needed in the release you use for experiements I'd move them to a profile specific to that for building the release with experiment related deps that are included in the release.


cmeiklejohn:
Critical to our experiments, we need to be able to build a rel and ship it to Amazon / Google, etc.  

How does this change the behavior when a release gets built?

tsloughter:
The release could still be built with those deps. I'll update the release to include those apps.

cmeiklejohn:
Can you branch off the membership branch for this?

tsloughter:
Sure. Looks like I have a number of other changes, like with xref, to get done as well.

tsloughter:
I need to figure out a good way to have it start the webmachine worker when the release runs but not otherwise.

tsloughter:
Also, `lasp_marathon_simulations` is used in a number of modules that seem to be core to functionality, while it (at least based on the module name) should be moved out to `simulations/`.

cmeiklejohn:
Can you retarget this for unstable?

cmeiklejohn:
I'm nervous to merge this until I'm able to run some experiments on Google Cloud to verify.  I'll try to do that over the next day or so.

cmeiklejohn:
Gah, can you handle this conflict?

cmeiklejohn:
Merging to run experiments.

PR #219, opened 2017-02-21
Enforce-once invariants

cmeiklejohn:


cmeiklejohn:
cc: @tsloughter

src/lasp_membership.erl
```diff
@@ -80,19 +80,19 @@ init([]) ->
     {reply, term(), #state{}}.
 
 handle_call(Msg, _From, State) ->
-    _ = lager:warning("Unhandled messages: ~p", [Msg]),
+    _ = lager:warning("Unhandled call: ~p", [Msg]),
     {reply, ok, State}.
```
tsloughter: Bit of a nitpick, since this has no call's it isn't going to turn into an issue here, but I've seen this cause bugs to go unnoticed and become very difficult to debug: do a `noreply` here instead of `reply`.

cmeiklejohn: Good idea.

cmeiklejohn: I'll fix it in the other PR.

PR #222, opened 2017-02-25
Divergence simulation.

cmeiklejohn:


cmeiklejohn:
When trying to capture divergence, I'm not precisely sure how I can know when to stop the simulation and count the divergence, because technically, I think I need to wait until all nodes perform anti-entropy with all other nodes?

src/lasp_divergence_client.erl
```diff
@@ -0,0 +1,207 @@
...
+            ok
+    end.
+
+%% @private
+max_events() ->
+    lasp_config:get(max_events, ?MAX_EVENTS_DEFAULT).
+
+%% @private
+max_events_reached(Events) ->
+    max_events() == Events orelse lasp_config:get(events_generated, false).
```
vitorenesduarte: `lasp_config:get(events_generated, false)`
should be enough, assuming the enforce function will run

PR #225, opened 2017-03-09
Remove EVENT_INTERVAL

vitorenesduarte:
- __L__: latency (operation time)
- __E__: event interval
- __B__: batch size
- __T__: total batch time
  - __T__ = (__B__ * __L__) + ((__B__ - 1) * __E__)

With this we can compute:
- latency:
  - __L__ = (__T__ - ((__B__  - 1) * __E__)) / __B__
- throughput:
  - __B__ / __T__

What we were logging was __T__ (since `?EVENT_INTERVAL = 0`).
So we need to change the log-processing script to compute latency correctly.


cmeiklejohn:
This should go into unstable, and no master.

PR #231, opened 2017-03-30
Experiments.

cmeiklejohn:


cmeiklejohn:
@vitorenesduarte Can you merge the latest unstable into this branch?

bin/kube-divergence-simulation
```diff
@@ -1,16 +1,17 @@
 #!/usr/bin/env bash
 
-CLIENTS=(1 2 4 6 8 10)
+CLIENTS=(1 2 4 1 2 4)
```
cmeiklejohn: I don't think this will work right, because of the labeling.  Does it?

vitorenesduarte:
@cmeiklejohn Yes. And I still need to fix that absolute path when copying the logs the `priv/evaluation`.

bin/kube-divergence-simulation
```diff
@@ -178,22 +181,23 @@ kubectl get pods
...
 fi
 
+echo "REDIS_SERVICE_HOST: ${REDIS_SERVICE_HOST}"
+echo "REDIS_SERVICE_PORT: ${REDIS_SERVICE_PORT}"
+
 echo "Downloading data from Redis."
 ./sync-redis.erl
-mkdir -p ~/Documents/lasp/priv/evaluation/logs
-cp -Rp /tmp/logs/* ~/Documents/lasp/priv/evaluation/logs
+mkdir -p ~/lasp/priv/evaluation/logs
```
cmeiklejohn: Can we parameterize this?  We both have different paths to where the lasp source code is.

cmeiklejohn:
Can we move the docker image to: https://hub.docker.com/u/lasplang so we can both push it?

cmeiklejohn:
I think we need to make the jitter percentage configurable through the environment, so we do not need to keep rebuilding this bootler image.

bin/kube-divergence-simulation
```diff
@@ -1,22 +1,24 @@
 #!/usr/bin/env bash
 
-# CLIENTS=(1 2 4 6 8 10)
-CLIENTS=(1)
-NODES=3
+CLIENTS=(1 2 4 1 2 4)
```
cmeiklejohn: Doing this is a problem that invalidates the results: this means that there are two nodes that are labeled for a run of client_number 4: because of the way the constraints are configured on launching containers, this **does not guarantee** that all of the containers for a single experiment are on the same node, therefore invalidating all of the results.

cmeiklejohn: cc: @vitorenesduarte

cmeiklejohn:
@tsloughter any objection to merging this into master?

tsloughter:
@cmeiklejohn no objection. Looks like the only code change is setting the jitter.

cmeiklejohn:
sounds good.

PR #237, opened 2017-05-06
remove github pages files

tsloughter:
Now that you are using readme.io I figured the files could be cleaned up.

Curious if `RIAK.md` and `EQC_CI_LICENCE.txt` are actually needed though?


cmeiklejohn:
No, you can remove them as well.

cmeiklejohn:
Please leave the images, but move them somewhere else.  This is where people go to get the assets.

tsloughter:
OK, I added back the images. Not sure of a better place to put them then where they are though.

cmeiklejohn:
Hrm, isn't that dtl file required for the internal webmachine interface?

cmeiklejohn:
It looks like things that are used for the homepage and things that are required for the webmachine interface are conflated in this PR, please make sure this is one removing the stuff related to the deprecated homepage.

tsloughter:
Oh, hm, I thought anything related to the webmachine interface was under `priv/`.

I can't find any reference to the index dtl file or use of a dtl template renderer.

PR #244, opened 2017-08-18
add erlang 20 for test

getong:


cmeiklejohn:
I think we need to amend this commit to include fixes to ensure that if we build on Erlang 20, Lasp actually builds.  Do you mind fixing the compilation errors related to ```export_all```?

cmeiklejohn:
It looks like this is still failing.  Also, don't you need to actual export the functions that were being exported by the `export_all`?

getong:
@cmeiklejohn 
The error msg:
```
===> Warning: lasp_core:reply_to_all/3 calls deprecated function gen_fsm:send_event/2 (Xref)
gen_fsm.erl:229: Warning: gen_fsm:send_event/2 is deprecated function (Xref)
make: *** [xref] Error 1
```
It seems that, we need to use gen_statem to replace the gen_fsm in erlang 20.
And erlang 19 build without error, the test code is ok.

cmeiklejohn:
Yes, these things need to be fixed before we can move to Erlang 20.  I didn't realize there were any outstanding issues, however, because I thought @tsloughter said that he was using Erlang 20.

cmeiklejohn:
Closing, we can't address this until the code compiles under Erlang 20.

PR #246, opened 2017-08-30
Bump rand_compat library.

cmeiklejohn:


cmeiklejohn:
@tsloughter Would you want to release a new version so Erleans can pull that dependency?

PR #248, opened 2017-09-19
version bump and upgrade partisan

tsloughter:


cmeiklejohn:
This is failing.

tsloughter:
Ah yea, partisan api changed. I can look at it later, but maybe @lrascao can see quicker exactly what the modifications needed in lasp are?

lrascao:
the biggest breaking change and the one that seems to be causing the failed tests is the new join api, an example can be found here: https://github.com/lrascao/plumtree/commit/a23fca6563d2a8440d8fc4587d4bc2bedb770c3d

cmeiklejohn:
@tsloughter please push a new release to hex when you have time, thanks!

PR #258, opened 2017-11-11
Use cast message, upgrade partisan.

cmeiklejohn:


cmeiklejohn:
Once this is merged, we need to bump the Lasp PG and Erleans dependencies on Lasp.

PR #269, opened 2017-11-20
handle gen_flow process crashing while waiting for response in single fire function

tsloughter:


cmeiklejohn:
looks good

cmeiklejohn:
```
18:43:04.252 [error] gen_server lasp_state_based_synchronization_backend terminated with reason: no match of right hand value {error,normal} in lasp_core:receive_value/2 line 982
```

tsloughter:
@cmeiklejohn are there multiple processes going on in gen flow? This issue doesn't make any sense if the process that sends the response is the same one that is shutting down normally (causing the `{error,normal}` response).

cmeiklejohn:
Weird.

Internally, gen_flow is going to use more than one process -- but, shouldn't this only be getting DOWN messages for the monitored process (or will it get DOWN for any spawned by the monitored process?)

cmeiklejohn:
Can the monitor be more specific?  For example, `{'DOWN', MonRef, process, _, Reason}` instead of `{'DOWN', _Ref, process, _, Reason}` ?

tsloughter:
It can, but I believe the issue is that the process it is monitoring is getting its down message to it before the process that sends the actual response is able to. Which wouldn't have been an issue if the response message was sent from the same process as the one being monitored.

cmeiklejohn:
It's bizarre: I'm going through the code and it appears that it should be the same process.

cmeiklejohn:
https://github.com/lasp-lang/gen_flow/blob/master/src/gen_flow.erl#L174
https://github.com/lasp-lang/lasp/blob/master/src/lasp_process.erl#L178
https://github.com/lasp-lang/lasp/blob/master/src/lasp_process.erl#L133

tsloughter:
If that is the case I'm more confused since it means that even in these cases the gen_flow process is exiting without a response first being sent.

tsloughter:
Is this maybe not just exposing the bug in a more visible way? I don't know what the failing tests were doing but could they have passed even though something didn't complete but now it was bubbling up to cause a crash and exposing it?

cmeiklejohn:
#270

PR #271, opened 2017-11-23
Remove os_mon.

cmeiklejohn:
@tsloughter is there a way to make this application conditionally start for some nodes so I don't need to remove it -- having trouble with it working in some environments.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -119,7 +119,7 @@
     ]
 }.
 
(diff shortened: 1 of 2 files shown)


tsloughter:
Not built in. You'd want to add it to your release as `{osmon, load}` and then have code in lasp's startup that starts it depending on the environment.

tsloughter:
+1 to this being removed too :), I'd been meaning to ask for it to be removed.

PR #272, opened 2017-11-24
Full mesh partial replication protocol.

cmeiklejohn:
- [x] Track interest groups, allow interest and disinterest.
- [x] Filter updates based on interest groups.
- [x] ObjectFilterFun everywhere must be a two arity call and take the metadata.
- [x] Object metadata should allow specification of which groups it belongs in.
- [ ] Because of transitive delivery, adapt, because some nodes will rely on transitivity to ensure convergence, which means a simple 1-1 exchange will only work under full replication.
- [x] Ensure functionality under partial replication.


cmeiklejohn:
Will address transitive work in a subsequent commit.

PR #288, opened 2018-03-19
Prevent process leakage.

cmeiklejohn:
Avoid a situtation where single fire processes terminate after receiving the response and stay registered to receive updates on thresholds firing.  Avoids a situtation where the process stays registered and allows the ETS table to grow indefinitely, ultimately crashing the node.


cmeiklejohn:
cc: @tsloughter

tsloughter:
Wooo! I'll give this a try.

cmeiklejohn:
@tsloughter is this good?

PR #304, opened 2019-12-27
upgrade partisan to 4.1.0 and remove unneeded time/rand_compat

tsloughter:
`time_compat` and `rand_compat` are no longer needed because the min OTP is 19.


tsloughter:
Updated with commit to upgrade plumtree to latest release, 0.6.0.

PR #311, opened 2020-12-31
Update README.md to mention erlang cookie

darkyne:
I initially faced an error when trying to cluster remote nodes together because machines did not share the same erlang.cookie file.
Ended up finding the problem and it went fine after putting the same erlang.cookie on the different machines.
This is just a suggestion but it would probably not hurt to mention it in the readme (especially in the case of beginners who are not used to erlang and distributed systems).

Happy new year 2021 by the way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Go to node b and retrieve the content of the CRDT:
 3> sets:to_list(AwMapSet).
 % [#{what => i_am_an_awmap_value}]
 ```
+If you try to make a cluster with remote nodes and face an error related to connection attempts from disallowed nodes, you may simply need to put the same erlang.cookie on the machines beforehand to allow them to communicate.
 
 ## Running a shell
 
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
This only happens if you're using the distributed Erlang topology and not one of the non-disterl ones.  It might be worth updating the README to reflect that, if you don't mind.

darkyne:
To be totally honest, I'm not really sure to understand the different kind of topologies (distributed erlang or not).

Personally, as a first approach, I cloned the repo and directly followed the little instructions from the readme file to create a little cluster, thus probably using the default configuration. 
Worked perfectly fine locally (using 127.0.0.1) so I wanted to try it on two computers (by following the exact same instructions but using their Ip address instead of 127.0.0.1 with two terminals, one locally and one on a remote computer via ssh). Got that error about disallowed node which was solved by putting the same erlang.cookie on both computers. This was mainly for testing purpose but was maybe not the correct approach.

Still, since I noticed erlang.cookie was mentioned nowhere, I thought it could be mentioned but I'm not sure exactly in which scenario it is required or not.

Probably not a real issue anyway :)

cmeiklejohn:
closing in favor of #315

